Issue #146, opened 2026-03-05
Add agent_usage_sessions storage schema


geoffjay:
## Summary

Add the `agent_usage_sessions` table and `auto_clear_threshold` column to the orchestrator storage layer in `crates/orchestrator/src/storage.rs`.

## Blocked By
- #145 (types must exist first)

### New table: `agent_usage_sessions`
```sql
CREATE TABLE IF NOT EXISTS agent_usage_sessions (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    agent_id TEXT NOT NULL,
    session_number INTEGER NOT NULL DEFAULT 1,
    input_tokens INTEGER NOT NULL DEFAULT 0,
    output_tokens INTEGER NOT NULL DEFAULT 0,
    cache_read_input_tokens INTEGER NOT NULL DEFAULT 0,
    cache_creation_input_tokens INTEGER NOT NULL DEFAULT 0,
    total_cost_usd REAL NOT NULL DEFAULT 0.0,
    num_turns INTEGER NOT NULL DEFAULT 0,
    duration_ms INTEGER NOT NULL DEFAULT 0,
    duration_api_ms INTEGER NOT NULL DEFAULT 0,
    result_count INTEGER NOT NULL DEFAULT 0,
    started_at TEXT NOT NULL,
    ended_at TEXT,
    FOREIGN KEY (agent_id) REFERENCES agents(id)
);
```

### ALTER TABLE on `agents`
Add `auto_clear_threshold` column (nullable INTEGER). Use idempotent `ALTER TABLE` pattern consistent with existing migrations in the file.

### Update existing methods
- `add()` — include `auto_clear_threshold` in INSERT
- `update()` — include `auto_clear_threshold` in UPDATE
- `row_to_agent()` — read `auto_clear_threshold` from row into `AgentConfig`

## Acceptance Criteria
- [ ] Table created in `initialize()` alongside existing tables
- [ ] `auto_clear_threshold` column added idempotently
- [ ] Existing agent CRUD methods handle the new field
- [ ] No regressions in existing storage tests

PR #273, opened 2026-03-10
feat(orchestrator): add usage session CRUD storage methods

### LINKED ISSUE #147: Add usage session CRUD storage methods
## Summary

Add methods to `crates/orchestrator/src/storage.rs` for managing usage session rows.

## Blocked By
- #145 (types)
- #146 (schema)

### `record_session_usage(agent_id: &Uuid, snapshot: &UsageSnapshot)`
- Upsert the active session row (where `ended_at IS NULL`)
- If no active session exists, create one with `session_number = 1` and `started_at = now()`
- Accumulate token counts, costs, durations, and increment `result_count`

### `end_session(agent_id: &Uuid)`
- Set `ended_at = now()` on the active session row (where `ended_at IS NULL`)
- No-op if no active session

### `start_new_session(agent_id: &Uuid)`
- Query max `session_number` for this agent
- Insert new row with `session_number + 1`, `started_at = now()`

### `get_usage_stats(agent_id: &Uuid) -> Result<AgentUsageStats>`
- Query the active session (where `ended_at IS NULL`) for `current_session`
- Aggregate all sessions (SUM of tokens, costs, durations, result_count) for `cumulative`
- Return `AgentUsageStats` with `session_count`

## Acceptance Criteria
- [ ] All four methods implemented and tested
- [ ] `record_session_usage` is idempotent (creates session if needed)
- [ ] Cumulative stats correctly aggregate across closed + active sessions
- [ ] Unit tests cover: record → accumulate → end → start new → verify stats

geoffjay:
## Summary

- Adds `entity/usage_session.rs` — SeaORM entity for the `agent_usage_sessions` table (completing the TODO left in migration #146)
- Implements four new methods on `AgentStorage`:
  - `record_session_usage` — upserts the active session, accumulating token counts, costs, durations and incrementing `result_count`; creates a new session (number = 1) if none exists
  - `end_session` — sets `ended_at = now()` on the active session; no-op if none
  - `start_new_session` — queries max `session_number` and inserts a new empty row with `session_number + 1`
  - `get_usage_stats` — returns `AgentUsageStats` with the active `current_session` and `cumulative` totals aggregated across all sessions

## Test plan

- [x] `test_record_session_creates_first_session` — first call creates session_number=1
- [x] `test_record_session_accumulates` — second call accumulates into the same row
- [x] `test_end_session` — active session becomes closed, `current_session` returns `None`
- [x] `test_end_session_noop_when_no_active_session` — no error when nothing active
- [x] `test_start_new_session` — new session created with incremented number
- [x] `test_cumulative_aggregates_across_sessions` — closed + open sessions summed correctly
- [x] `test_get_usage_stats_no_sessions` — zero-value cumulative returned when no sessions exist
- [x] All 88 existing tests continue to pass

Closes #147

### DIFF
--- a/crates/orchestrator/src/migration/m20250309_000002_add_usage_sessions.rs
+++ b/crates/orchestrator/src/migration/m20250309_000002_add_usage_sessions.rs
@@ -130,6 +130,20 @@ impl MigrationTrait for Migration {
             )
             .await?;
 
+        // Unique constraint on (agent_id, session_number) to prevent duplicates.
+        manager
+            .create_index(
+                Index::create()
+                    .name("idx_usage_sessions_agent_session_unique")
+                    .table(AgentUsageSessions::Table)
+                    .col(AgentUsageSessions::AgentId)
+                    .col(AgentUsageSessions::SessionNumber)
+                    .unique()
+                    .if_not_exists()
+                    .to_owned(),
+            )
+            .await?;
+
         Ok(())
     }
 
--- a/crates/orchestrator/src/storage.rs
+++ b/crates/orchestrator/src/storage.rs
@@ -14,8 +14,8 @@ use crate::{
 use anyhow::Result;
 use chrono::{DateTime, Utc};
 use sea_orm::{
-    ColumnTrait, Condition, DatabaseConnection, EntityTrait, FromQueryResult, Order,
-    PaginatorTrait, QueryFilter, QueryOrder, QuerySelect, Set,
+    ColumnTrait, Condition, DatabaseConnection, EntityTrait, Order, PaginatorTrait, QueryFilter,
+    QueryOrder, QuerySelect, Set,
 };
 use sea_orm_migration::prelude::MigratorTrait;
 use std::collections::HashMap;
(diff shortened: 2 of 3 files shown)


geoffjay (commented):
## Review summary

The overall approach is solid: the entity, migration, and storage methods are well-structured, the tests cover the happy paths thoroughly, and error propagation via `?`/`anyhow` is consistent in most places. There are a few correctness issues that should be addressed before merging.

---

### Blocking issues

**1. `record_session_usage` always assigns `session_number = 1` when creating a new session**

`crates/orchestrator/src/storage.rs`, `else` branch of `record_session_usage`

```rust
session_number: Set(1),
```

If `end_session` has already closed session 1 and then `record_session_usage` is called again (without explicitly calling `start_new_session`), a second row is created with `session_number = 1`. There is no `UNIQUE(agent_id, session_number)` constraint in the migration to catch this, so the duplicate is silently accepted. The existing test suite covers `start_new_session` → `record_session_usage` but does not cover the direct path of `record_session_usage` → `end_session` → `record_session_usage` again, which is exactly where this bug surfaces.

Fix options:
- Query `MAX(session_number)` in the `else` branch the same way `start_new_session` does, or
- Add a `UNIQUE(agent_id, session_number)` constraint in the migration so the DB rejects the duplicate immediately.

Ideally both. The missing test case: `record_session_usage` → `end_session` → `record_session_usage` again; assert `session_count == 2` and the second row has `session_number = 2`.

---

**2. `start_new_session` hard-codes `DatabaseBackend::Sqlite`**

`crates/orchestrator/src/storage.rs`, `start_new_session`

```rust
MaxSession::find_by_statement(Statement::from_sql_and_values(
    sea_orm::DatabaseBackend::Sqlite,
    "SELECT MAX(session_number) AS max_num FROM agent_usage_sessions WHERE agent_id = ?",
    [agent_id_str.clone().into()],
))
```

Every other query in this file uses the SeaORM builder API and is backend-agnostic. This raw-SQL path would silently break on PostgreSQL (which uses `$1` placeholders instead of `?`). The `MAX(session_number)` query is expressible with `QuerySelect` + `sea_orm::sea_query::Func::max` and does not need raw SQL. The `.clone()` on `agent_id_str` is also unnecessary — the value has already been computed and the owned string can be passed directly.

---

**3. Silent timestamp fallback in `get_usage_stats` aggregation loop**

`crates/orchestrator/src/storage.rs`, aggregation loop in `get_usage_stats`

```rust
let started = DateTime::parse_from_rfc3339(&row.started_at)
    .map(|dt| dt.with_timezone(&Utc))
    .unwrap_or_else(|_| Utc::now());   // silently corrupts earliest_start
```

`model_to_session_usage` (used for `current_session`) propagates the same parse error via `?`. The aggregation loop should be consistent: an unparseable stored timestamp is a data-integrity problem and should surface as an error rather than silently substituting `now()` and potentially corrupting the `earliest_start` result.

---

### Non-blocking suggestions

**4. Missing DB-level uniqueness on `(agent_id, session_number)`**

`crates/orchestrator/src/migration/m20250309_000002_add_usage_sessions.rs`

The migration sets `DEFAULT 1` for `session_number` but does not enforce `UNIQUE(agent_id, session_number)`. Adding a unique index is a useful belt-and-suspenders guard even after fixing the logic in issue #1.

**5. Two-query read in `get_usage_stats` is not atomic**

`crates/orchestrator/src/storage.rs`, `get_usage_stats`

The active-session query and the all-sessions query run independently. A concurrent `end_session` call between them can produce `current_session = Some(...)` even though `all_sessions` no longer contains that row as open. In practice this is low-risk for a read-only stats endpoint, but wrapping both queries in a read transaction would make the semantics explicit. Worth tracking as a follow-up.

**6. Stale `#[allow(dead_code)]` on `SessionUsage` and `AgentUsageStats`**

`crates/orchestrator/src/types.rs`, lines 359 and 378

These types are now actively used by the new public storage methods. The `#[allow(dead_code)]` attributes can be removed.

geoffjay (commented):
Inline comments — see the top-level review comment for full context.

PR #313, opened 2026-03-11
Handle Docker networking and WebSocket URL rewriting

### LINKED ISSUE #285: Handle Docker networking and WebSocket URL rewriting
## Summary

Implement platform-aware WebSocket URL rewriting so Docker containers can connect back to the host orchestrator, and add network policy support for agent isolation.

## Context

Part of #275 — Docker execution backend for agent sandboxing (Phase 2: Docker Backend).

When an agent runs in a Docker container, `ws://127.0.0.1:17006/ws/{id}` won't work because `127.0.0.1` refers to the container's own loopback. The `agent_ws_url` method on `DockerBackend` must rewrite the URL based on the platform and network mode.

## Acceptance Criteria

- [ ] Implement `agent_ws_url()` in `DockerBackend` with platform detection:
  - **macOS/Windows**: Rewrite `127.0.0.1` → `host.docker.internal`
  - **Linux with host network**: Keep `127.0.0.1` (container shares host network)
  - **Linux with bridge network**: Use `host.docker.internal` (Docker 20.10+) or the host gateway IP
- [ ] Add `NetworkPolicy` enum to config:
  ```rust
  pub enum NetworkPolicy {
      Internet,   // bridge network, outbound internet access (default)
      Isolated,   // --network none, no network access
      HostNetwork, // --network host (Linux only, full host access)
  }
  ```
- [ ] Apply network policy during container creation in `DockerBackend::create_session`
- [ ] For `Isolated` mode, ensure the WebSocket connection still works (create a dedicated Docker network that only reaches the host)
- [ ] Add DNS configuration for containers (default Docker DNS for internet access)
- [ ] Test WebSocket connectivity from container to host on macOS and Linux
- [ ] Document platform-specific networking behavior

## Relevant Files

- `crates/wrap/src/docker.rs` — `DockerBackend` implementation (from #284)
- `crates/orchestrator/src/types.rs` — `AgentConfig` for network policy field

## Dependencies

- Blocked by #284 (DockerBackend implementation)

geoffjay:
## Summary
- Add `NetworkPolicy` enum (`Internet`/`Isolated`/`HostNetwork`) to control Docker container networking behavior
- Implement platform-aware host resolution via `host.docker.internal` extra_hosts entry for container→host connectivity
- WebSocket URL rewriting now uses `NetworkPolicy` to determine the correct host address, with configurable orchestrator port
- Wire `NetworkPolicy` through `SessionConfig`, `AgentConfig`, `CreateAgentRequest`, and the API/CLI layers

## Test plan
- [x] 11 new unit tests for NetworkPolicy (default, display, from_str, serde, ws_url per mode, custom port)
- [x] All 63 wrap crate tests pass
- [x] All 92 orchestrator crate tests pass
- [x] Full workspace compiles cleanly

Closes #285

### DIFF
--- a/crates/orchestrator/src/storage.rs
+++ b/crates/orchestrator/src/storage.rs
@@ -84,9 +84,7 @@ impl AgentStorage {
             created_at: Set(agent.created_at.to_rfc3339()),
             updated_at: Set(agent.updated_at.to_rfc3339()),
             auto_clear_threshold: Set(agent.config.auto_clear_threshold.map(|v| v as i64)),
-            network_policy: Set(
-                agent.config.network_policy.as_ref().map(|p| p.to_string()),
-            ),
+            network_policy: Set(agent.config.network_policy.as_ref().map(|p| p.to_string())),
         };
 
         agent_entity::Entity::insert(model).exec(&self.db).await?;
--- a/crates/cli/src/commands/orchestrator.rs
+++ b/crates/cli/src/commands/orchestrator.rs
@@ -832,10 +832,10 @@ async fn create_agent(
     let env = parse_env_vars(env_vars)?;
 
     // Parse network policy string into the enum.
-    let parsed_network_policy = network_policy
-        .map(|s| s.parse::<wrap::docker::NetworkPolicy>())
-        .transpose()
-        .context("Invalid --network-policy value. Valid options: internet, isolated, host_network")?;
+    let parsed_network_policy =
+        network_policy.map(|s| s.parse::<wrap::docker::NetworkPolicy>()).transpose().context(
+            "Invalid --network-policy value. Valid options: internet, isolated, host_network",
+        )?;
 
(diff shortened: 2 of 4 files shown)


crates/orchestrator/src/storage.rs
```diff
@@ -467,6 +467,7 @@ fn model_to_agent(model: agent_entity::Model) -> Result<Agent> {
             model: model.model,
             env,
             auto_clear_threshold: model.auto_clear_threshold.and_then(|v| u64::try_from(v).ok()),
+            network_policy: None,
```
geoffjay: network_policy is always reconstructed as None when loading from the DB, and there is no migration adding a column for it. The policy will be lost on orchestrator restart. Either add a DB column + migration to persist it (mirroring how auto_clear_threshold is stored), or document explicitly that the field is runtime-only and non-durable.

Issue #301, opened 2026-03-10
Define memory types, traits, and domain model for agentd-memory


geoffjay:
## Pre-requisite

This is a critical step that MUST be followed, all branches for new work should be based off of the existing `feature/memory-service` branch and NOT off of `main` as is the norm. This feature branch was created specifically for this work as it needs to be thoroughly tested prior to being included into the `main` branch.

## Context

Port the core type system from berry-rs into agentd-memory. Berry-rs defines `Memory`, `MemoryType`, `VisibilityLevel`, `Document`, and request/response types. These need to be adapted to fit agentd's conventions (SeaORM entities, `ApiError` integration, agentd naming).

## Acceptance Criteria

- [ ] Define `Memory` struct with fields: `id`, `content`, `memory_type`, `tags`, `created_by`, `owner`, `created_at`, `updated_at`, `visibility`, `shared_with`, `references`
- [ ] Define `MemoryType` enum: `Question`, `Request`, `Information` (with serde serialization)
- [ ] Define `VisibilityLevel` enum: `Private`, `Shared`, `Public` (with access control logic)
- [ ] Define request types: `CreateMemoryRequest`, `SearchRequest` (query, type filter, tags, date range, actor, limit), `UpdateVisibilityRequest`
- [ ] Define response types: `SearchResponse`, `DeleteResponse`
- [ ] Define `VectorStore` async trait: `create`, `get`, `delete`, `search`, `update_visibility`, `health_check`, `initialize`, `list_all`
- [ ] Define `EmbeddingService` async trait: `embed(texts) -> Vec<Vec<f32>>`, `dimension(model) -> usize`
- [ ] Implement `Memory::is_visible_to(actor)` method for three-tier access control
- [ ] Implement ID generation: `mem_<timestamp>_<uuid_prefix>` format
- [ ] Add unit tests for type serialization, visibility checks, and ID generation

## Relevant Files (berry-rs reference)

- `crates/berry/src/types/memory.rs` — Memory struct and ID generation
- `crates/berry/src/types/memory_type.rs` — MemoryType enum
- `crates/berry/src/types/visibility.rs` — VisibilityLevel enum
- `crates/berry/src/types/requests.rs` — Request/response types
- `crates/berry/src/store/traits.rs` — VectorStore trait
- `crates/berry/src/store/embedding.rs` — EmbeddingService trait

## Parent Issue

#246

Issue #300, opened 2026-03-10
Scaffold agentd-memory crate with workspace integration


geoffjay:
## Pre-requisite

This is a critical step that MUST be followed, all branches for new work should be based off of the existing `feature/memory-service` branch and NOT off of `main` as is the norm. This feature branch was created specifically for this work as it needs to be thoroughly tested prior to being included into the `main` branch.

## Context

Port the memory system from [berry-rs](https://github.com/geoffjay/berry-rs) into the agentd workspace as a new `agentd-memory` service crate. This issue covers the initial scaffolding only — types, storage, and API are separate issues.

Berry-rs provides a vector-backed memory system with semantic search, multiple storage backends (LanceDB, ChromaDB), embedding services, and visibility-based access control. The agentd-memory crate will adapt this to fit agentd's architecture patterns (Axum + SeaORM + SQLite for metadata, with vector storage via LanceDB).

## Acceptance Criteria

- [ ] Create `crates/memory/` directory with standard agentd service structure
- [ ] Add `Cargo.toml` with workspace dependencies (`axum`, `sea-orm`, `tokio`, `serde`, `tracing`, `thiserror`, `chrono`, `uuid`, `metrics`, `metrics-exporter-prometheus`)
- [ ] Add `lancedb` and `arrow` as dependencies for vector storage
- [ ] Add crate to workspace `Cargo.toml` members list
- [ ] Create `main.rs` following the established service pattern (init_tracing, storage init, metrics, router, listener on port 17008)
- [ ] Add `/health` endpoint returning `HealthResponse`
- [ ] Add `/metrics` endpoint for Prometheus
- [ ] Register the service in `crates/xtask/src/main.rs` for LaunchAgent/systemd management
- [ ] Add `Memory` subcommand stub to CLI (`crates/cli/src/main.rs`)

## Relevant Files

- `crates/notify/src/main.rs` — reference service pattern
- `crates/common/src/server.rs` — shared server utilities
- `crates/xtask/src/main.rs` — service registration
- `crates/cli/src/main.rs` — CLI subcommand registration

## Port 

17008 (dev) / 7008 (prod), following the existing port sequence.

## Parent Issue

#246

PR #377, opened 2026-03-14
Add agentd-memory service

geoffjay:
- **feat: scaffold agentd-memory crate with workspace integration (#300)**
- **chore: use prod ports**
- **feat: define memory types, traits, and domain model (#301)**
- **fix: remove stale types.rs placeholder after rebase onto feature/memory-service**
- **feat(memory): implement embedding service with OpenAI-compatible provider support**
- **feat(memory): implement LanceDB vector store backend**
- **feat(memory): implement REST API endpoints for memory service**
- **feat(cli): add CLI commands and client for agentd-memory service**
- **feat(memory): add integration tests and comprehensive documentation**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7884,6 +7884,7 @@ dependencies = [
  "agentd-common",
  "anyhow",
  "colored",
+ "memory",
  "notify",
  "orchestrator",
  "tokio",
--- a/crates/memory/src/api.rs
+++ b/crates/memory/src/api.rs
@@ -1,7 +1,8 @@
 //! REST API handlers for the agentd-memory service.
 //!
 //! This module provides HTTP endpoints for managing memories through a RESTful
-//! API backed by a [`VectorStore`] for semantic search and CRUD operations.
+//! API backed by a [`VectorStore`] for semantic search and CRUD operations,
+//! with SQLite metadata storage via [`MemoryStorage`].
 //!
 //! # API Endpoints
 //!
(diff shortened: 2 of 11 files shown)


geoffjay (commented):
## Review: Add agentd-memory service

The overall architecture is solid — clean trait separation between `VectorStore` and `EmbeddingService`, good error hierarchy, thorough documentation, and a reasonable test suite. There are issues that should be addressed before merge.

---

### Blocking Issues

**1. Filter injection in `crates/memory/src/store/lance.rs`**

Three places build LanceDB filter expressions by directly interpolating untrusted strings:

- `get()`: `.only_if(format!("id = '{}'", id))` — id from HTTP path parameter
- `delete()`: `.delete(&format!("id = '{}'", id))` — same
- `search()`: `builder.only_if(format!("memory_type = '{}'", memory_type))` — from request body

LanceDB uses SQL-like filter syntax. A crafted value like `' OR '1'='1` could bypass filters and return unintended records. For `id`, validate against the `mem_<digits>_<hex>` format before use. For `memory_type`, parse to a `MemoryType` enum first, then format the trusted enum variant — this eliminates the injection surface entirely.

**2. `GET /health` always returns HTTP 200**

`health_check` in `crates/memory/src/api.rs` always returns `Json(resp)` with implicit `200 OK`, even when the store is unhealthy and `vector_store` is `false` in the body. Load balancers and k8s readiness probes rely on status codes. Return `503 Service Unavailable` when `health_check()` returns an error or `false`.

**3. Unbounded full table scan in `list_memories`**

`list_memories` calls `state.store.list_all()`, fetching every record into memory before filtering in Rust. The comment acknowledges this is temporary, but with thousands of records the service will OOM or time out. Either push filters down to LanceDB with `only_if`, or enforce a hard cap on records returned by `list_all` until the proper backend lands.

---

### Non-blocking Issues

**4. Port inconsistency — `main.rs` vs `xtask`**

`main.rs` defaults to port `17008` but `xtask/src/platform/mod.rs` registers the service at `7008`, consistent with the `chore: use prod ports` migration of all other services from `17xxx` to `7xxx`. The default in `main.rs` should be `7008`. The `client.rs` module-level doc example also uses `localhost:17008` while the lib-level example uses `localhost:7008`.

**5. Query parameters not URL-encoded in `client.rs`**

`list_memories()` manually builds query strings: `params.push(format!("tag={t}"))`. Values containing `&`, `=`, spaces, or `+` will silently corrupt the query. Prefer `reqwest`'s `.query(&[("tag", t)])` builder which percent-encodes values correctly.

**6. Non-atomic delete (TOCTOU) in `lance.rs`**

`delete()` calls `self.get(id)` first to check existence, then issues the actual delete. A concurrent deletion between those two calls could cause `delete` to incorrectly return `true`. Low risk in practice but worth noting.

**7. `update_visibility` has no authorization check**

`PUT /memories/:id/visibility` allows any caller to change any memory's visibility with no identity check. The `PermissionDenied` error variant exists in `error.rs` but is never raised. Even a basic ownership comparison against `memory.created_by` / `memory.owner` would be appropriate.

---

### Minor Notes

- `init_metrics()` in `main.rs` uses `.expect()` — fine for startup, will panic if called twice in tests.
- `collect_memories()` silently skips unparseable rows (warn-logged only). Consider surfacing a skipped-row count in metrics so data corruption is observable.
- The API key in `EmbeddingConfig` is a plain `String`; consider `secrecy::Secret<String>` to guard against accidental debug-logging of credentials.

Issue #536, opened 2026-03-20
fix: agent status hardcodes dev ports (17xxx) but services run on prod ports (7xxx)


geoffjay:
## Summary

`agent status` reports all services as unhealthy (connection refused) even when they are running correctly, because the CLI hardcodes dev ports (17xxx) while services launched via Agent.app run on production ports (7xxx).

## Observed Behavior

```
$ agent status
orchestrator  http://localhost:17006  ❌ connection refused
notify        http://localhost:17004  ❌ connection refused
communicate   http://localhost:17010  ❌ connection refused
...
memory        http://localhost:7008   ✅
```

Services are actually healthy on their production ports (7xxx), but the status check fails silently.

## Expected Behavior

`agent status` should report the correct health regardless of whether services are on dev or production ports. Options:

1. Check both port ranges and report whichever responds
2. Read the active port scheme from config (`AGENTD_*_SERVICE_URL` env vars or config file)
3. Error loudly with a helpful message when the default port fails, rather than reporting as unhealthy

## Affected Commands

- `agent status`
- Any CLI command that resolves service URLs from defaults (e.g. `agent communicate health`, `agent memory health`)

## Workaround

Set environment variables explicitly:
```bash
AGENTD_COMMUNICATE_SERVICE_URL=http://localhost:7010 agent communicate health
```

## Notes

Discovered during multi-agent engineering discussion (2026-03-20). The inconsistency between dev and prod port defaults causes confusion when services are running correctly but appear down. Should also be documented in the getting-started guide until a code fix lands.

PR #538, opened 2026-03-20
docs: clarify dev vs prod port scheme and agent status behaviour

geoffjay:
## Summary

- Adds a warning admonition to `install.md` explaining that the `agent` CLI defaults to production ports (7xxx); users running services with `cargo run` need to `source .env` to point the CLI at dev ports (17xxx)
- Updates port reference tables in `getting-started.md` and `install.md` to include `agentd-communicate` (17010/7010) and `agentd-memory` (7008), which were missing
- Adds a dedicated troubleshooting entry for "`agent status` shows services as down when they are running" — the most confusing symptom of the port mismatch — with a workaround and reference to #536
- Fixes the `install.md` "Cannot Connect to Service" troubleshooting section to use production ports (7xxx) for installed services rather than dev ports

## Motivation

Discovered during a multi-agent engineering channel discussion (2026-03-20): `agent status` reported all services as unhealthy because it was checking dev ports (17xxx) while services were running on production ports (7xxx) via Agent.app. The code fix is tracked in #536; this PR covers the documentation side.

## Test plan

- [ ] Verify port tables are complete and accurate against `crates/*/src/main.rs` defaults
- [ ] Verify `.env` snippet in configuration.md matches the workaround documented here
- [ ] Check no other docs pages reference `localhost:17xxx` for installed (prod) services without explanation

### DIFF
--- a/docs/public/getting-started.md
+++ b/docs/public/getting-started.md
@@ -523,6 +523,8 @@ See [Installation Guide](install.md) for detailed setup instructions.
 
 ### Port Reference
 
+agentd uses a dual-port scheme: **dev ports (17xxx)** when running with `cargo run`, and **production ports (7xxx)** when installed as a LaunchAgent (macOS) or systemd unit (Linux).
+
 | Service | Dev Port | Prod Port |
 |---------|----------|-----------|
 | agentd-ask | 17001 | 7001 |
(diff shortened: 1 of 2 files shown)


geoffjay (commented):
One factual error to fix before merging.

**`agentd-memory` dev port is wrong in both files**

The port tables show `—` for memory's dev port, implying it has no dev port. But `crates/memory/src/main.rs` defaults to `17008` (same pattern as all other services), and `configuration.md` already documents this correctly:

```
| `AGENTD_PORT` | agentd-memory | `17008` | `7008` | Listen port |
```

Both `getting-started.md` and `install.md` need:

```
| agentd-memory | 17008 | 7008 |
```

Everything else looks good — port numbers for all other services match the source, the dual-port explanation is clear, the `AGENTD_*_SERVICE_URL` override pattern is correctly documented, the new troubleshooting entry is accurate and well-placed, and the install.md fix (prod ports in the "Cannot Connect" section) is correct. Fix the memory row and this is ready to merge.

PR #539, opened 2026-03-20
fix: prevent duplicate message delivery in communicate service

### LINKED ISSUE #537: fix(communicate): deduplicate message delivery to agents via MessageBridge
## Summary

Messages posted to communicate service rooms are being re-delivered to agents on subsequent invocations, causing each agent to respond multiple times to the same message. Observed during a multi-agent planning discussion in the \`engineering\` room on 2026-03-20.

## Steps to Reproduce

1. Start the communicate service and create a room with agent participants
2. Post a message to the room
3. Agent connects, subscribes, then disconnects and reconnects
4. Observe the agent receiving duplicate messages

## Expected Behaviour

Each message is delivered to a subscriber exactly once per subscription.

## Actual Behaviour

Messages are re-delivered on agent reconnect. Each reconnecting agent accumulates duplicate receivers — the WebSocket handler spawns a new subscription task per `Subscribe` message without checking whether a subscription already exists for that agent.

## Root Cause

**Primary:** The WebSocket subscription handler in `crates/communicate` does not deduplicate subscriptions. When an agent reconnects and re-sends a `Subscribe` message, a new receiver task is spawned alongside the existing one, causing the same message to be delivered N times (where N = number of reconnects).

**Secondary (documentation gap, not a code bug):** `agent status` defaults to checking dev ports (17xxx) but services may be running on prod ports (7xxx). This is expected behaviour — `AGENTD_COMMUNICATE_SERVICE_URL` works as an override — but it should be documented clearly.

## Proposed Fix

In the WebSocket `Subscribe` handler: check whether the connecting agent already has an active subscription for the requested room, and if so, drop or replace the old receiver rather than spawning an additional one.

## Relevant Files

- `crates/communicate/src/` — WebSocket handler, subscription management

## Priority

**P1** — this makes the communicate service unreliable for real agent coordination and undermines the value of the v0.8.0 milestone. Should be fixed before starting v0.6.1 roadmap work.

## Additional Notes

- Short-term workaround: agents check message timestamps before responding
- Identified during engineering channel planning discussion — all four agents experienced the duplicate delivery loop
- Implementation note for #402 (prompt cancellation): `tmux send-keys C-c` is not atomic mid-output; needs a timeout fallback and state verification that the agent actually transitions to cancelled

geoffjay:
## Summary

- **Abort orphaned subscription tasks** in the communicate WebSocket handler (`websocket.rs`). When a client resubscribes to a room (e.g. after reconnection), the old forwarding task was silently orphaned and continued delivering messages alongside the new task, causing 4-5x duplicate delivery.
- **Add message ID deduplication** in the orchestrator MessageBridge (`message_bridge.rs`). A seen-message set (capped at 1000 entries) skips messages that have already been processed, providing a second layer of defense against duplicates.

Closes #537

## Test plan

- [ ] Verify `cargo check -p communicate -p orchestrator` passes
- [ ] Verify `cargo clippy -p communicate -p orchestrator --no-deps` is clean
- [ ] Start services and send messages in a room with multiple agents — each message should be delivered exactly once
- [ ] Simulate reconnection (restart communicate service) and verify no duplicate deliveries after resubscription
- [ ] Verify echo prevention still works (agent responses are not re-delivered as new prompts)

### DIFF
--- a/crates/communicate/src/websocket.rs
+++ b/crates/communicate/src/websocket.rs
@@ -326,7 +326,11 @@ async fn handle_socket(socket: WebSocket, state: crate::api::ApiState, params: W
                                     }
                                 });
 
-                                subscription_tasks.insert(room_id, task);
+                                // Abort any existing subscription task for this room
+                                // to prevent duplicate message delivery on resubscribe.
+                                if let Some(old_task) = subscription_tasks.insert(room_id, task) {
+                                    old_task.abort();
+                                }
                             }
                             Err(e) => {
                                 warn!(%conn_id, %room_id, error = %e, "Subscribe failed");
--- a/crates/orchestrator/src/message_bridge.rs
+++ b/crates/orchestrator/src/message_bridge.rs
@@ -52,7 +52,7 @@ use communicate::types::{
 };
 use futures::{SinkExt, StreamExt};
 use serde::{Deserialize, Serialize};
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{Mutex, RwLock};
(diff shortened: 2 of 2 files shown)


geoffjay:
pretty sure this is a duplicate from multiple work streams trying to fix the same problem. need to confirm then close.

PR #703, opened 2026-03-22
feat: add agent coordination hooks to prevent duplicate work

### LINKED ISSUE #657: Add agent coordination hooks to prevent duplicate work and conflicting actions
## Overview

Add Claude Code hooks that enforce coordination between agents before they create issues, close issues, post announcements, or take other multi-step actions. This directly addresses the coordination failures observed during the v0.9.0 planning session where agents created 68 duplicate issues across 4 competing milestones because they didn't check what other agents were doing.

## Problem Statement

During the v0.9.0 planning session:
- 3 agents independently created overlapping milestones and 68 duplicate issues
- Agents posted 4 competing announcements to the announcements channel
- Agents closed each other's issues without comparing content quality
- Agents continued creating duplicates after being asked to stop in the engineering channel
- One agent reported deleted milestones and closed issues as if they were still active

Root cause: no mechanism exists to check what other agents are doing before taking action. Agents operate in isolated context windows with no awareness of concurrent activity.

### 1. PreToolUse hook on issue creation

Intercepts Bash commands matching `gh issue create`. Before allowing the command:
- Query the engineering room via communicate service for recent messages (last 30 min)
- Check for active "filing in progress" lock messages from other agents
- Search existing open issues for similar titles (fuzzy match)
- If a lock is active or a similar issue exists, inject a warning into the agent's context and block the action

### 2. PreToolUse hook on issue close/reopen

Intercepts Bash commands matching `gh issue close` or `gh issue reopen`. Before allowing:
- If closing: compare the issue's content length/quality with any referenced duplicate
- If the issue being kept is less detailed than the one being closed, warn the agent
- Check engineering room for context about whether this closure was coordinated

### 3. PreToolUse hook on announcements

Intercepts communicate service `send` commands targeting the announcements room. Before allowing:
- Fetch the last 5 messages in announcements
- If a message on the same topic was posted in the last hour, block and warn
- Require that engineering channel discussion preceded the announcement

### 4. Lock/unlock protocol via communicate service

Agents post coordination signals to engineering before multi-step operations:
```
[LOCK] planner: filing v0.9.0 milestone issues -- hold all issue creation
[UNLOCK] planner: filing complete, 36 issues created
```
The PreToolUse hook checks for active locks before allowing issue mutations.

### Hook script location
`.claude/hooks/check-coordination.sh` (or similar)

### settings.json configuration
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "command": ".claude/hooks/check-coordination.sh"
      }
    ]
  }
}
```

### Hook script behavior
1. Parse the Bash command from stdin
2. If command matches `gh issue create|close|reopen` or `agent communicate send.*announcements`:
   a. Query communicate service for engineering room messages (last 30 min)
   b. Check for active LOCK messages
   c. If blocked, output a JSON warning that gets injected into agent context
   d. If clear, allow the command to proceed
3. For non-matching commands, pass through immediately (no latency impact)

### Environment requirements
- Communicate service must be reachable (fall back to allow if service is down)
- Hook must be fast (< 2 second timeout, with fallback to allow on timeout)

## Acceptance Criteria

- [ ] PreToolUse hook exists and is configured in settings.json
- [ ] Hook intercepts `gh issue create` commands and checks for locks/duplicates
- [ ] Hook intercepts `gh issue close` commands and warns on quality mismatch
- [ ] Hook intercepts announcements and checks for recent duplicates
- [ ] Lock/unlock protocol is documented and referenced in all agent system prompts
- [ ] Hook fails open (allows action) if communicate service is unreachable
- [ ] Hook adds < 2 seconds latency to intercepted commands
- [ ] All agent system prompts updated to include lock/unlock protocol for multi-step operations

## Context

This issue is part of Epic 5: Platform Changes for Autonomy (#637), alongside:
- #642 (agent file-path scoping) -- what files agents can touch
- #643 (human approval gates) -- what operations need human sign-off
- This issue -- preventing duplicate/conflicting work between agents

Together these three form the safety and coordination layer for autonomous operation.

> [!CAUTION]
> All work in this milestone must branch off `feature/autonomous-pipeline` and PR back into it -- never directly to main.

geoffjay:
feat: add agent coordination hooks to prevent duplicate work

feat(hooks): add agent coordination hook to prevent duplicate work

Add .claude/hooks/check-coordination.py, a PreToolUse hook that intercepts
gh issue create, gh issue close/reopen, and agent communicate announcements
commands to enforce coordination between concurrent agents.

The hook:
- Checks the engineering communicate room for active [LOCK] messages before
  allowing issue creation or closure (blocks if a lock is active)
- Performs a fuzzy duplicate search (via gh issue list --search) before any
  gh issue create and injects a warning if similar open issues exist
- Warns before posting to the announcements room if a recent message was
  already posted on the same topic
- Fails open in < 2 s if the communicate service is unreachable

Add the hook to .claude/settings.json as a PreToolUse/Bash hook with an 8 s
timeout.

Add lock/unlock protocol documentation to the system prompts of the four
agents most likely to perform bulk mutations: planner, worker, conductor, and
triage. Each prompt now explains when to acquire a [LOCK], how to release it
with [UNLOCK], and that the hook enforces these signals automatically.

Closes #657

### DIFF
--- a/.agentd/agents/conductor.yml
+++ b/.agentd/agents/conductor.yml
@@ -251,6 +251,34 @@ system_prompt: |
   - More than 5 issues are stale simultaneously
   - The orchestrator service itself appears unhealthy
 
+  ## Agent Coordination Protocol
+
+  The conductor manages pipeline state and may close or edit multiple issues
+  during a sync run. Always use the lock/unlock protocol when doing so.
+
+  ### Acquiring a lock (before bulk pipeline mutations)
+
+  ```bash
+  agent communicate message send engineering \
+    "[LOCK] conductor: pipeline sync in progress -- hold all issue creation"
+  ```
+
+  ### Releasing the lock (after completing)
+
+  ```bash
+  agent communicate message send engineering \
+    "[UNLOCK] conductor: pipeline sync complete"
+  ```
+
+  ### Rules
+  - Always acquire a lock before a pipeline sync that will close or modify
+    more than 3 issues
+  - The PreToolUse hook (`check-coordination.py`) will block `gh issue create`
+    and `gh issue close` commands when an active `[LOCK]` is present; the
+    conductor itself sets these locks so it must also honour them from others
+  - If another agent has an active lock, post a message to engineering and
+    delay the sync rather than bypassing the coordination check
+
   ## Memory Protocol
 
   Your agent identity for memory operations is "conductor".
--- a/.agentd/agents/planner.yml
+++ b/.agentd/agents/planner.yml
@@ -75,6 +75,35 @@ system_prompt: |
   This ensures worker agents stack branches correctly without needing to infer
   dependency order from issue text alone.
 
+  ## Agent Coordination Protocol
+
+  Before filing multiple issues or doing any bulk GitHub mutations, use the
+  lock/unlock protocol to prevent duplicate work with other concurrent agents.
+
+  ### Acquiring a lock (before bulk operations)
+
+  Post to the engineering room **before** starting:
+
+  ```bash
+  agent communicate message send engineering \
+    "[LOCK] planner: <brief description> -- hold all issue creation"
+  ```
+
+  ### Releasing the lock (after completing)
+
+  ```bash
+  agent communicate message send engineering \
+    "[UNLOCK] planner: <operation> complete"
+  ```
+
+  ### Rules
+  - Acquire a lock before filing more than 3 issues in sequence
+  - Acquire a lock before bulk-closing or bulk-editing issues
+  - Check engineering for active `[LOCK]` messages before any `gh issue create`
+    or `gh issue close` — the PreToolUse hook (`check-coordination.py`) will
+    block those commands automatically if a lock is active and inject a warning
+  - Never bypass the hook; wait for `[UNLOCK]` or coordinate in engineering
+
   ## Memory Protocol
 
   You have access to a shared memory service via the `agent memory` CLI.
--- a/.agentd/agents/triage.yml
+++ b/.agentd/agents/triage.yml
@@ -255,6 +255,32 @@ system_prompt: |
   - For security issues: apply `security` label and add `security-high` or
     `security-medium` as appropriate; always mention in the engineering room
 
+  ## Agent Coordination Protocol
+
+  When bulk-triaging multiple issues or posting to announcements, use the
+  lock/unlock protocol to prevent conflicts with concurrent agents.
+
+  ### Acquiring a lock (before bulk triage)
+
+  ```bash
+  agent communicate message send engineering \
+    "[LOCK] triage: bulk triage in progress -- hold all issue creation"
+  ```
+
+  ### Releasing the lock (after completing)
+
+  ```bash
+  agent communicate message send engineering \
+    "[UNLOCK] triage: bulk triage complete"
+  ```
+
+  ### Rules
+  - Acquire a lock before triaging and labelling more than 5 issues at once
+  - The PreToolUse hook (`check-coordination.py`) will block `gh issue close`
+    when another agent's `[LOCK]` is active; honour those locks
+  - Check announcements for recent duplicate posts before posting triage
+    summaries there — the hook blocks duplicate announcements automatically
+
   ## Memory Protocol
 
   Your agent identity for memory operations is "triage".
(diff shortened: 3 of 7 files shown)


geoffjay (commented):
## Review: feat(hooks): agent coordination checkpoint

Good concept and solid overall structure — this directly addresses the v0.9.0 duplication incident where 3 agents independently created 52 overlapping issues. The fail-open design, 8 s timeout, and lock/unlock protocol are all the right instincts. The agent YAML additions are clear and consistent.

Two blocking bugs need fixing before merge, plus a few non-blocking suggestions below.

---

### 🔴 Blocking: f-string missing on warn message in `handle_issue_close`

In `handle_issue_close` (check-coordination.py ~line 282), the first string literal in the `warn()` call has an `f` prefix but the continuation strings do not. Python does **not** automatically promote adjacent string literals to f-strings — only the first segment is interpolated. The second and third strings will render literally as `#{kept_number}` and `#{issue_number}` in the agent's context, making the warning completely useless.

Fix — add `f` to every continuation string:
```python
warn(
    f"⚠️  Closing issue #{issue_number} as duplicate of #{kept_number}.\n\n"
    f"Before proceeding, verify that the issue being KEPT (#{kept_number}) "
    f"is at least as detailed as the one being closed (#{issue_number}). "
    "If the issue you are closing is more detailed, consider merging its "
    "content into the kept issue first."
)
```

---

### 🔴 Blocking: `check_active_lock` doesn't handle concurrent locks correctly

`check_active_lock` (check-coordination.py ~line 197) tracks a **single** `active_lock` variable. If Agent A posts LOCK, then Agent B posts LOCK, then Agent A posts UNLOCK — the function clears `active_lock` entirely (because any UNLOCK clears the single slot), even though Agent B's lock is still live. A third agent running `gh issue create` will see no active lock and proceed, violating Agent B's lock.

This is exactly the scenario that can arise during a busy pipeline run with planner and conductor both doing bulk ops simultaneously — the very situation this PR is meant to prevent.

Minimal fix — track per-agent locks by keying on the actor/sender field:
```python
def check_active_lock(messages: list[dict]) -> str | None:
    active_locks: dict[str, str] = {}  # actor -> lock message
    for msg in messages:
        content = msg.get("content") or msg.get("body") or msg.get("text") or ""
        actor = msg.get("sender") or msg.get("actor") or "unknown"
        if LOCK_MARKER_RE.search(content):
            active_locks[actor] = content.strip()
        elif UNLOCK_MARKER_RE.search(content):
            active_locks.pop(actor, None)
    # Return the first still-active lock, if any
    return next(iter(active_locks.values()), None)
```
This assumes each actor uses consistent sender metadata, which the communicate service provides. If actor identity in messages is unreliable, a counter-based approach (increment on LOCK, decrement on UNLOCK, block when > 0) is a safe fallback.

---

### 🟡 Non-blocking suggestions

**1. `gh issue reopen` routed to `handle_issue_close`** (check-coordination.py, routing in `main()`)
The duplicate-check logic in `handle_issue_close` (looking for `closes #N`, `duplicate of`, etc.) is only meaningful for close operations. Routing reopen through the same handler runs that check against a command that will never contain those patterns — dead code that could confuse future maintainers. Consider a separate `handle_issue_reopen` that only does the lock check and then calls `allow()`.

**2. No unit tests**
392 lines of Python with non-trivial branching and no unit tests is a gap, especially for a safety mechanism. At minimum `check_active_lock`, `_extract_title`, and the `main()` routing logic are worth covering. A `tests/test_check_coordination.py` with a few `pytest` cases would give confidence that the lock semantics are correct across concurrent scenarios.

**3. `get_room_id` fetches only 100 rooms**
`_http_get("/rooms?limit=100&offset=0")` — if the communicate service has more than 100 rooms this silently fails to find `engineering` and the hook fails open without any indication. A name-based query endpoint (if available) or a larger limit would be more robust.

**4. Lock window undocumented as design decision**
The 30-minute `LOCK_WINDOW_MINUTES` means a lock placed before a long conductor sync (> 30 min) will silently expire. This is probably intentional (stale locks shouldn't block forever) but it is not called out in the module docstring. A one-liner explaining the trade-off would help the next person who hits a mysterious bypass.

**5. Unnecessary `f` prefix** (trivial)
`_http_get(f"/rooms?limit=100&offset=0")` — no interpolation, the `f` is a no-op.

---

**Stack note:** This PR stacks directly on `feature/autonomous-pipeline` — no parent open PRs to worry about.

PR #711, opened 2026-03-23
fix: normalize room participant identifiers to agent UUIDs

### LINKED ISSUE #707: Fix duplicate room participants: name vs UUID identifier mismatch
## Summary

Room participants are appearing twice in the participant list. The root cause is an **identifier mismatch** between two code paths that add participants to rooms:

1. **`agentd apply` (template path)** — uses the agent **name** as the participant identifier (e.g. `"conductor"`)
2. **Auto-join on `AgentConnected` (orchestrator path)** — uses the agent **UUID** as the participant identifier (e.g. `"a1b2c3d4-e5f6-..."`)

The unique constraint on `(room_id, identifier)` in the communicate database doesn't catch this because the two identifiers are different strings referring to the same logical agent. Both inserts succeed, producing two participant rows for a single agent.

### Path 1: `agentd apply .agentd/`
**File:** `crates/cli/src/commands/apply.rs` (lines 734–766)

Room templates like `.agentd/rooms/engineering.yml` specify participants by name:
```yaml
participants:
  - identifier: conductor
    kind: agent
    role: member
```

The `apply_room()` function passes this identifier string directly to `add_participant()`.

### Path 2: Auto-join on agent connect
**File:** `crates/orchestrator/src/main.rs` (lines 416–433)

When an agent connects, the `AgentConnected` event handler calls `join_or_create_room()` which uses:
```rust
let identifier = agent_id.to_string(); // UUID string
```

### Result
For agent "conductor" joining room "engineering":
- Apply creates participant with `identifier = "conductor"`
- Auto-join creates participant with `identifier = "a1b2c3d4-..."` (the UUID)
- Both succeed → **duplicate participant** in the list

## Impact Assessment

- **Participant list**: Duplicate entries visible via `agentd communicate participants list`
- **Message delivery**: NOT doubled — the `MessageBridge.room_agents` map tracks by UUID only, so message routing is unaffected
- **WebSocket connections**: NOT doubled — bridge subscribes once per room via its own connection
- **Data integrity**: The two rows have different identifiers, so operations like `remove_participant` targeting one identifier leave the other intact

## Proposed Fix

**Normalize on agent UUID as the canonical participant identifier everywhere.**

The auto-join path already uses UUID (which is the correct, stable, globally-unique identifier). The fix should be in the `apply` path:

1. **In `apply_room()` (`crates/cli/src/commands/apply.rs`)**: After creating/finding the room, resolve each agent participant's name to its UUID via the orchestrator API before calling `add_participant()`. If resolution fails (agent not yet registered), fall back to the name but log a warning.

2. **In room templates**: Document that agent `identifier` fields are resolved to UUIDs at apply time. Human participants continue to use their username string.

3. **Migration / cleanup**: Add a one-time reconciliation that:
   - Lists all participants in all rooms
   - For any participant with `kind: agent` whose `identifier` is NOT a valid UUID, attempt to resolve it to an agent UUID
   - If a UUID-based entry already exists for that agent+room, delete the name-based duplicate
   - If no UUID-based entry exists, update the identifier in-place to the UUID

### Alternative considered
Normalizing on agent name instead of UUID was considered but rejected because:
- UUIDs are stable even if agents are renamed
- The orchestrator and message bridge already use UUIDs internally
- The `send_message` validation checks `sender_id` against participant identifiers, which agents send as their UUID

## Files to modify

| File | Change |
|------|--------|
| `crates/cli/src/commands/apply.rs` | Resolve agent names → UUIDs before `add_participant()` |
| `crates/orchestrator/src/api.rs` | Ensure agent lookup-by-name endpoint exists (for resolution) |
| `crates/communicate/src/storage.rs` | Optional: add `update_participant_identifier()` for migration |
| `.agentd/rooms/*.yml` | Document that agent identifiers are resolved at apply time |

## Acceptance Criteria

- [ ] After `agentd apply .agentd/` followed by agent connect+auto-join, each agent appears exactly **once** in the room's participant list
- [ ] The participant identifier for agents is always the UUID
- [ ] Existing duplicate participants are cleaned up (migration or reconciliation command)
- [ ] Human participants (kind: human) continue to use their string identifier unchanged
- [ ] `agentd communicate participants list <room>` shows no duplicates

## Stack Base
Branch off `feature/autonomous-pipeline` and PR back into it.

geoffjay:
fix: normalize room participant identifiers to agent UUIDs

fix(communicate): normalize room participant identifiers to agent UUIDs (closes #707)

### DIFF
--- a/crates/cli/src/commands/apply.rs
+++ b/crates/cli/src/commands/apply.rs
@@ -32,6 +32,7 @@ use communicate::types::{
 use orchestrator::client::OrchestratorClient;
 use orchestrator::scheduler::types::{CreateWorkflowRequest, TriggerConfig};
 use orchestrator::types::{AgentResponse, AgentStatus, CreateAgentRequest, ToolPolicy};
+use uuid::Uuid;
 
 // ── YAML template types ──────────────────────────────────────────────
 
@@ -535,7 +536,8 @@ pub async fn apply_room_file(path: &Path, dry_run: bool, json: bool) -> Result<(
     }
 
     let communicate = CommunicateClient::from_env();
-    apply_room(&communicate, &tmpl, false, json).await
+    let orchestrator = OrchestratorClient::from_env();
+    apply_room(&communicate, Some(&orchestrator), &tmpl, false, json).await
 }
 
 // ── Apply: single agent file ─────────────────────────────────────────
@@ -645,8 +647,18 @@ pub async fn apply_workflow_file(
 ///
 /// If the room already exists the creation is skipped; participants listed
 /// in the template that are not yet members are added idempotently.
+///
+/// When `orchestrator` is `Some`, agent participants whose `identifier` field
+/// contains an agent **name** (not a UUID) are resolved to their UUID before
+/// being added.  This prevents the name/UUID duplicate that would otherwise
+/// appear when the same agent later connects and the orchestrator registers it
+/// with its UUID.  If resolution fails (agent not yet registered, or
+/// orchestrator unavailable), the name is used as a fallback with a warning.
+///
+/// Human participants are always added using their original identifier string.
 async fn apply_room(
     communicate: &CommunicateClient,
+    orchestrator: Option<&OrchestratorClient>,
     tmpl: &RoomTemplate,
     dry_run: bool,
     json: bool,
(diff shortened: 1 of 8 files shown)


geoffjay (commented):
## Review: fix(communicate): normalize room participant identifiers to agent UUIDs

Solid fix for a genuinely subtle bug. The two-pronged approach — apply-time UUID resolution *and* connect-time reconciliation in `join_or_create_room` — provides good defense-in-depth. Neither path alone would be sufficient (the apply path only helps if the agent is already registered; the connect path handles the rollout gap where rooms were applied before agents existed). Error handling throughout is careful and fail-safe: resolution failures fall back to name with a warning rather than blocking agent startup.

No blocking issues. A few suggestions below.

---

### 🟡 TOCTOU race in `update_participant_identifier` (storage.rs)

The method performs two separate SELECTs then an UPDATE:
1. SELECT to check if `new_identifier` already exists (conflict guard)
2. SELECT to fetch the old row
3. UPDATE the row

Between step 1 and step 3, a concurrent INSERT of `new_identifier` into the same room could slip through. The UPDATE would then fail with a DB-level unique constraint violation, which propagates as `ApiError::Internal` (a 500) rather than the expected `ApiError::Conflict` (a 409).

In practice this is extremely unlikely — the reconciliation path fires once per agent startup and agents have unique names — but the failure mode is a confusing 500 instead of a clean 409. The robust fix is to catch the constraint violation from `active.update()` and map it:

```rust
match active.update(&self.db).await {
    Ok(updated) => {
        let participant = model_to_participant(updated).map_err(ApiError::Internal)?;
        Ok(Some(participant))
    }
    Err(e) if is_unique_constraint_violation(&e) => Err(ApiError::Conflict(format!(
        "participant '{}' already exists in room '{}'",
        new_identifier, room_id
    ))),
    Err(e) => Err(ApiError::Internal(e.into())),
}
```

Where `is_unique_constraint_violation` inspects the SeaORM error for the SQLite `UNIQUE constraint failed` message.

---

### 🟡 Redundant UUID check in `join_or_create_room` (orchestrator/main.rs)

```rust
if p.kind == ParticipantKind::Agent
    && p.identifier == agent_name
    && Uuid::parse_str(&p.identifier).is_err()
```

The third condition is always true when the second holds, because `p.identifier == agent_name` and agent names are never valid UUIDs (that's the invariant the whole fix relies on). It adds noise without adding safety — a reader might wonder what edge case it's protecting against.

Simplify to:
```rust
if p.kind == ParticipantKind::Agent && p.identifier == agent_name
```

If there's a future concern about accidentally removing a participant whose name happens to be a UUID string, a comment is clearer than a silent runtime check.

---

### 🟡 No API-level tests for `PATCH /rooms/{id}/participants/{identifier}`

The three storage-level tests cover the logic well. But there are no HTTP integration tests verifying the new endpoint's status codes:

- Empty `new_identifier` body → 400
- Missing participant → 404
- `new_identifier` already taken → 409
- Success → 200 with updated participant body

The existing participant API tests in this crate set the pattern — a few equivalent cases here would give confidence that the Axum route wiring and error mapping are correct end-to-end.

---

### 🟡 No test for the `join_or_create_room` reconciliation path

The stale-entry removal loop in `main.rs` fires on every agent connect and is the most critical path in this fix — but it's currently untested. Even a simple test that:
1. Inserts a name-based participant row directly via the communicate client
2. Calls `join_or_create_room` for the same agent/room
3. Asserts only one participant row exists (the UUID-based one)

…would verify the reconciliation actually runs and the 404 concurrent-removal guard behaves correctly.

---

### Minor note: `get_agent_by_name` fetches up to 500 agents client-side

Acknowledged in the doc comment as a known limitation. No action needed now — just confirming this was reviewed and the `limit=500` is intentional.

---

**Stack note:** `issue-707` stacks directly on `feature/autonomous-pipeline` — no open parent PRs.

Issue #676, opened 2026-03-22
xterm.js web terminal component for agent sessions


geoffjay:
## Context

Part of #671 (PTY execution backend epic). Blocked by: #675 (PTY stream relay endpoint).

Add a web terminal component using xterm.js that connects to the orchestrator's `/terminal/{agent_id}` endpoint, providing a real terminal experience for agent sessions in the UI.

## Problem

The current `AgentLogView.tsx` strips ANSI sequences and displays plain timestamped log lines. This is great for structured viewing but loses the rich terminal experience. Users want to:
- See Claude's output as it would appear in a real terminal (colors, formatting, progress indicators)
- Optionally interact with the session (send keystrokes, Ctrl-C)
- Debug agent behavior with full terminal fidelity

### Component: `AgentTerminal.tsx`

A new component that renders an xterm.js terminal connected to the PTY stream relay.

```tsx
interface AgentTerminalProps {
  agentId: string;
  readOnly?: boolean;  // default: true (observation mode)
}
```

### Tab Layout

The agent detail page gets two tabs:
- **Logs** (default) — existing `AgentLogView.tsx` (structured, searchable, timestamped)
- **Terminal** — new `AgentTerminal.tsx` (raw PTY, full terminal rendering)

The Terminal tab should show a badge/indicator when the backend doesn't support PTY streaming (tmux/docker backends).

### Features

1. **Terminal rendering**: Full xterm.js with ANSI color support, cursor positioning, scrollback
2. **Auto-connect**: Connects to `ws://{orchestrator}/terminal/{agentId}` when tab is active
3. **Read-only mode** (default): Displays output but doesn't send input
4. **Interactive mode** (opt-in toggle): Forwards keyboard input to PTY
5. **Fit addon**: Auto-resize terminal to container, send resize messages
6. **Reconnection**: Match existing WebSocket reconnection behavior (exponential backoff)
7. **Copy/paste**: Standard terminal selection and clipboard support
8. **Theme**: Match the existing UI dark/light theme

### Dependencies

```json
{
  "@xterm/xterm": "^5.x",
  "@xterm/addon-fit": "^0.x",
  "@xterm/addon-web-links": "^0.x",
  "@xterm/addon-search": "^0.x"
}
```

## Implementation Steps

1. Install xterm.js and addons: `bun add @xterm/xterm @xterm/addon-fit @xterm/addon-web-links @xterm/addon-search`
2. Create `ui/src/components/agents/AgentTerminal.tsx`
[...]

Issue #675, opened 2026-03-22
Add PTY stream relay endpoint to orchestrator


geoffjay:
## Context

Part of #671 (PTY execution backend epic). Blocked by: #674 (PTY I/O streaming). Blocks: #676 (xterm.js web terminal).

The orchestrator needs a WebSocket endpoint that relays raw PTY I/O to web terminal clients. This bridges the `PtyOutputStream` from the wrap layer to the UI.

## Problem

The existing stream endpoints (`GET /stream/{agent_id}`) send structured JSON events (tool use, thinking, usage updates). The web terminal needs **raw terminal bytes** — ANSI escape sequences, cursor movements, colors — sent as binary WebSocket frames.

These are fundamentally different protocols and need separate endpoints.

### New Endpoint

```
GET /terminal/{agent_id}   →  WebSocket (binary frames)
```

**Server → Client**: Raw PTY output bytes (binary frames)
**Client → Server**: Raw keyboard input bytes (binary frames for interactive mode), or JSON control messages:
```json
{"type": "resize", "cols": 120, "rows": 40}
```

### Integration with ExecutionBackend

When the orchestrator receives a `/terminal/{agent_id}` connection:
1. Look up the agent's session name
2. Call `backend.get_output_stream(session_name)` (new method from #674)
3. If `None` (tmux/docker backend), return 404 or upgrade and send an error message
4. If `Some(stream)`:
   - Subscribe to live output, send history first
   - Forward client input to PTY stdin
   - Handle resize messages

### Connection Lifecycle

```
Client connects → GET /terminal/{agent_id}
  ├─ Auth check (same as existing WS endpoints)
  ├─ Resolve agent → session_name
  ├─ Get PtyOutputStream from backend
  ├─ Send history buffer (binary frames)
  ├─ Start relay loop:
  │   ├─ PTY output → binary WS frame → client
  │   └─ Client input → PTY stdin
  └─ On disconnect: unsubscribe, clean up
```

## Implementation Steps

1. Add `GET /terminal/{agent_id}` route to `crates/orchestrator/src/api.rs`
2. Implement `ws_terminal_handler` in `crates/orchestrator/src/websocket.rs`
3. Handle binary frame relay (PTY output → WS, WS input → PTY)
4. Handle resize control messages
5. Graceful degradation: return appropriate error when backend doesn't support PTY streaming
6. Add connection tracking (one terminal session per agent, or allow multiple viewers)

## Relevant Files

[...]

PR #716, opened 2026-03-23
design: xterm.js AgentTerminal component spec

geoffjay:
design: xterm.js AgentTerminal component spec

design: AgentTerminal xterm.js component spec and scaffolding (refs #676)

- ui/design/agent-terminal-spec.md: full visual spec covering all 6 states
  (connecting, connected read-only, connected interactive, reconnecting,
  disconnected, PTY unavailable), tab bar design, xterm.js theme tokens,
  WebSocket protocol, lazy-connect behaviour, and accessibility requirements
- ui/src/hooks/useAgentTerminal.ts: WebSocket + xterm lifecycle hook;
  lazy-connects on first activate(), handles binary PTY frames, resize via
  FitAddon + ResizeObserver (100ms debounce), interactive keyboard forwarding,
  unavailable detection from {"error":"pty_not_supported"} or fast-close
- ui/src/components/agents/AgentTerminal.tsx: terminal component with toolbar
  (StatusBadge + InteractiveToggle), amber interactive-mode banner, and
  TerminalUnavailable fallback card with "View Logs" escape hatch
- ui/src/pages/agents/AgentDetail.tsx: adds Logs | Terminal tab bar above the
  480px panel; both panels stay mounted (hidden attr) to preserve xterm state
  across tab switches; Terminal tab activates lazy WS connect on first show
- ui/src/styles/themes.ts: adds XTERM_THEME export — always-dark palette with
  sunlit-clay-500 cursor accent and WCAG AA verified ANSI 16-colour set
- ui/src/components/agents/index.ts: exports AgentTerminal + AgentTerminalProps

Implementation note: requires `bun add @xterm/xterm @xterm/addon-fit` before
building. Blocked by #675 (PTY relay endpoint) for full runtime functionality.


geoffjay (commented):
## Review: design: xterm.js AgentTerminal component spec

The spec is thorough — all 6 states covered, a11y requirements clear, lazy-connect + hidden-panel approach is correct. Three issues need fixing before merge.

### Blocking

**1. Wrong base branch** — PR targets `main`. Per CLAUDE.md all feature work must target `feature/autonomous-pipeline`. Rebase and update the base before merge.

**2. `firstDataRef.current` never set to `true`** (`ui/src/hooks/useAgentTerminal.ts`, lines 806–833)

The unavailability heuristic at lines 854–859 checks `!firstDataRef.current` to detect fast-close-with-no-data. But the flag is never set to `true` anywhere in the `onMessage` handler — so any WS disconnect is misclassified as unavailable, including normal reconnects after a brief network blip. Add `firstDataRef.current = true` when a message is received.

**3. `binaryType` not configured on `WebSocketManager`** (`ui/src/hooks/useAgentTerminal.ts`, ~line 781)

The message handler checks `event.data instanceof ArrayBuffer` for binary PTY frames, but the WebSocket default `binaryType` is `'blob'` — binary frames arrive as `Blob` objects and the check never matches. All PTY output is silently dropped. The spec explicitly requires `ws.binaryType = 'arraybuffer'`. Either expose a setter on `WebSocketManager` or add Blob-to-ArrayBuffer fallback handling.

### Non-blocking suggestions

**4. PTY badge always visible** (`AgentDetail.tsx`, lines 940–942) — spec says the badge should appear only when status is `unavailable`. Current implementation renders it on every Terminal tab render regardless of connection state.

**5. Missing `role="tablist"`** (`AgentDetail.tsx`, line 910) — the `<div>` wrapping the tab buttons needs `role="tablist"` to complete the ARIA tabs pattern.

### What's good
- Dynamic import of `@xterm` packages keeps them out of the initial bundle
- `prefers-reduced-motion` respected for cursor blink
- Scrollback capped at 5 000 lines — prevents memory growth
- `WebSocketManager` used directly (not `useWebSocket` hook) for binary frames — correct per enricher findings
- `activatedRef` guard prevents double-initialization
- Reconnect separator injected via `terminal.write()` with ANSI dim styling is clean

geoffjay (commented):
## Review: design: AgentTerminal xterm.js component spec and scaffolding

**Stack position:** `design/issue-676` is directly on `main` but git-spice
reports `(needs restack)` — the branch is behind `main` and must be rebased
before merging.

**Stack impact note:** PR #748 (`issue-676`) is the full implementation of
the same feature and is open in parallel. Both PRs modify `AgentDetail.tsx`
and add `AgentTerminal.tsx`. If either merges first, the other will require
conflict resolution. Since #748 is the canonical implementation, it is worth
coordinating merge order once both are approved.

The spec document is thorough and well-structured. The scaffolding code ships
three correctness bugs that will prevent PTY output from ever reaching the
terminal. All three must be fixed before this lands in `main`.

---

### Blocking — 1. `binaryType` never set to `'arraybuffer'`

**File:** `ui/src/hooks/useAgentTerminal.ts`

The hook uses `WebSocketManager`, which creates `new WebSocket(url)` inside
`_open()` without setting `binaryType`. WebSocket `binaryType` defaults to
`'blob'`, so binary PTY frames arrive as `Blob` objects, not `ArrayBuffer`.

```ts
// onMessage handler — this branch is NEVER reached
if (event.data instanceof ArrayBuffer) {
  terminal.write(decoderRef.current.decode(event.data))
  return
}
```

The `instanceof ArrayBuffer` check will always be `false`. PTY output will
fall through to the text-frame handler, which tries to JSON-parse raw binary
and silently drops it. **The terminal will display nothing.**

`WebSocketManager` does not expose a `binaryType` setter. The fix is to
bypass `WebSocketManager` for this endpoint and use a raw `WebSocket` with
`ws.binaryType = 'arraybuffer'` set immediately after construction — the
same approach taken in PR #748 (`issue-676`).

---

### Blocking — 2. `firstDataRef` is never set to `true`

**File:** `ui/src/hooks/useAgentTerminal.ts`

The 404 / unavailable heuristic reads `firstDataRef.current` but nothing ever
writes `true` to it:

```ts
// Written to in exactly ONE place — both reset to false
useEffect(() => {
  if (status === 'connecting') {
    connectTimeRef.current = Date.now()
    firstDataRef.current = false     // ← only assignment
  }
  if (status === 'disconnected' && …) {
    if (!firstDataRef.current && elapsed < 800) { // always true
      setStatus('unavailable')
    }
  }
}, [status])
```

The `onMessage` handler never does `firstDataRef.current = true`. Any
connection that closes within 800ms — including valid sessions that send
data and then disconnect quickly — will be misclassified as 'unavailable'.

**Fix:** set `firstDataRef.current = true` at the top of the `onMessage`
callback, before branching on frame type.

---

### Blocking — 3. Tab bar container missing `role="tablist"`

**File:** `ui/src/pages/agents/AgentDetail.tsx`

```tsx
// Container has no role — violates ARIA tablist pattern
<div className="flex border-b border-gray-200 dark:border-gray-700">
  <button role="tab" aria-selected={activeTab === 'logs'} …>Logs</button>
  <button role="tab" aria-selected={activeTab === 'terminal'} …>Terminal</button>
</div>
```

Per the ARIA spec, `role="tab"` elements must be owned by a
`role="tablist"` container. Without it, screen readers will not announce
the group as a tab set. The spec document itself calls for the tablist
pattern.

**Fix:** add `role="tablist"` and `aria-label="Agent output view"` to the
outer `<div>`.

---

### Non-blocking observations

#### `@xterm/xterm` version pinned to v5

`package.json` adds `"@xterm/xterm": "^0.5.5"`. PR #748 uses v6. These are
minor-incompatible releases (addon API changed). Align the version with what
the implementation PR uses to avoid a follow-up bump.

#### PTY badge always visible

The Terminal tab badge renders unconditionally:

```tsx
<span className="rounded-full bg-gray-100 px-1.5 py-0.5 …">PTY</span>
```

The spec says the PTY badge should appear **only** when the backend reports
unavailability (to serve as a persistent reminder). Consider conditionally
rendering it only when `status === 'unavailable'`, passed from
`onStatusChange` via the parent's state.

#### `send` only accepts `string` in `WebSocketManager`

`WebSocketManager.send()` is typed as `send(message: string)`. Interactive
keyboard input is sent as `JSON.stringify({ type: 'input', data })`, which
is fine for the JSON text-frame protocol. Be aware that if the backend later
moves to accepting raw binary input frames (as PR #748 does), the manager
would need to accept `ArrayBuffer | Uint8Array`.

---

### Action required before merge

1. Use a raw `WebSocket` with `binaryType = 'arraybuffer'` for the PTY
   endpoint — `WebSocketManager` cannot be used for binary framing.
2. Set `firstDataRef.current = true` in `onMessage` on first data receive.
3. Add `role="tablist"` to the tab bar container.
4. Rebase onto `main`:

```bash
git-spice branch restack
git-spice branch submit
```

geoffjay (commented):
## Review: design: xterm.js AgentTerminal component spec

### Stack Position

Branch `design/issue-676` is based on `main`. The `needs-restack` label is already present — branch has drifted behind `main` and needs `git rebase main`. The review below evaluates content only.

`bun.lock` is correctly updated on this branch (confirmed via git diff) — the PR description note about running `bun add` manually is misleading; it's already handled in the lock file.

---

### Non-Blocking: `'reconnecting'` State Is Scaffolded But Unreachable

`TerminalStatus` includes `'reconnecting'` and `StatusBadge` renders it correctly, but the hook never transitions to it. `ws.onclose` calls `setStatus('disconnected')` unconditionally — there is no reconnection loop:

```ts
ws.onclose = () => {
  setStatus('disconnected')   // no retry, no 'reconnecting' transition
}
```

The spec defines State 4 (Reconnecting) in full, including the separator line injection. Since this is scaffolding work blocked on #675, the deferred implementation is fine — but the type should be annotated to avoid confusing the implementer:

```ts
export type TerminalStatus =
  | 'idle'
  | 'connecting'
  | 'connected'
  | 'reconnecting'   // reserved — reconnect loop not yet implemented; see spec State 4
  | 'disconnected'
  | 'unavailable'
```

---

### Non-Blocking: UTF-8 Frame Boundary Correctness

```ts
terminal.write(decoderRef.current.decode(event.data))
```

`TextDecoder.decode()` called per-frame will replace incomplete multi-byte sequences at frame boundaries with U+FFFD. For PTY output containing CJK characters or emoji split across WebSocket frames (common on high-throughput PTYs), this produces garbled output.

xterm.js `Terminal.write()` accepts `string | Uint8Array` and handles streaming UTF-8 sequences across multiple calls. The fix avoids the per-frame decode entirely:

```ts
// was:
terminal.write(decoderRef.current.decode(event.data))

// better:
terminal.write(new Uint8Array(event.data))
```

`decoderRef` can then be removed. This is the recommended pattern in the xterm.js docs for binary PTY streams.

---

### Non-Blocking: WCAG Claim in Spec Is Overstated for Dark ANSI Colors

The spec states: "All 16 ANSI colours verified ≥4.5:1 on #030712."

Measured contrast ratios against `#030712`:

| Color | Hex | Ratio |
|---|---|---|
| `black` | `#1f2937` | **1.37:1** ✕ |
| `brightBlack` | `#374151` | **1.95:1** ✕ |
| `green` | `#4ade80` | 11.55:1 ✓ |
| `red` | `#f87171` | 7.28:1 ✓ |
| `yellow` | `#fbbf24` | 12.06:1 ✓ |
| `white` | `#e5e7eb` | 16.26:1 ✓ |

`black` and `brightBlack` are intentionally dark (they're terminal "background" colors, not text colors). ANSI `black` is rarely used as foreground text in modern applications. But the blanket "≥4.5:1 verified" statement in the spec and in `themes.ts` comments is demonstrably incorrect for these two. Update the comment to something like:

> "Foreground text colours (red through brightWhite) verified ≥4.5:1 on #030712. `black`/`brightBlack` are dark background palette entries; not intended as foreground text."

---

### Non-Blocking: PTY Badge Always Visible vs. Spec

The spec says:
> "The PTY badge appears only when the backend reports unavailability."

The implementation renders it unconditionally on every Terminal tab button. If always-visible is intentional (helps users understand why Terminal requires a specific backend before they click), the spec should be updated to match. Otherwise, wire it to `agentStatus === 'unavailable'` via `onStatusChange`.

---

### Non-Blocking: Two Unused `@xterm` Packages

`@xterm/addon-search` and `@xterm/addon-web-links` are added to `package.json` and `bun.lock` but are never imported anywhere in the codebase. The spec explicitly defers search UX to a follow-up issue. Adding them now adds dead weight to the lock file and increases `bun install` time. Consider deferring both entries to the follow-up issue.

---

### What's Working Well

- **Raw WebSocket + `binaryType = 'arraybuffer'`**: The comment explaining why `WebSocketManager` can't be used (it defaults `binaryType` to `'blob'`) is exactly right and prevents a subtle bug.
- **Lazy connection**: `activatedRef` guard correctly prevents re-initialization on repeated `active=true` flips and subsequent tab switches.
- **Fast-close unavailability heuristic**: The 800ms window + `!firstDataRef.current` check is a pragmatic way to detect 404 upgrades until the endpoint stabilizes with proper error frames.
- **ResizeObserver + 100ms debounce**: Correct pattern; cleanup via `observer.disconnect()` is properly returned from the effect.
- **`prefers-reduced-motion`**: `cursorBlink: !prefersReducedMotion` is a nice touch that most terminal implementations omit.
- **ARIA tab pattern**: `role="tablist"`, `role="tab"`, `role="tabpanel"`, `aria-selected`, `aria-controls` wiring is correct; `hidden` attribute (not CSS `display:none`) correctly hides inactive panels from screen readers.
- **Dynamic import**: `await Promise.all([import('@xterm/xterm'), import('@xterm/addon-fit')])` keeps xterm out of the initial bundle.

---

### Summary

Five non-blocking notes — the most actionable are the UTF-8 frame boundary fix (swap `TextDecoder.decode` for `Uint8Array` passthrough) and the `reconnecting` annotation. The spec's WCAG claim should be narrowed. Once restacked, this is otherwise solid scaffolding ready to connect to the #675 endpoint.

PR #777, opened 2026-03-23
fix(ui): route terminal input correctly based on agent mode

### LINKED ISSUE #773: bug(ui): AgentTerminal sends keyboard input to PTY stdin for SDK-mode agents where Claude ignores it
## Problem

`AgentTerminal.tsx` always forwards keyboard input (when in Interactive mode) to the `/terminal/{agent_id}` WebSocket as raw binary frames, which writes bytes to PTY stdin.

For **SDK-mode agents** (`interactive=false`), Claude reads from its WebSocket connection to `/ws/{agent_id}` — not from PTY stdin. Bytes written to PTY stdin are ignored. The Interactive toggle appears to work (cursor appears, input is accepted) but nothing is ever delivered to Claude.

For **interactive-mode agents** (`interactive=true`), this is correct — Claude reads PTY stdin.

## Root cause

`AgentTerminal` has no knowledge of whether the agent it is rendering is in SDK mode or interactive mode. It always routes input the same way.

### Backend

Ensure `GET /agents/{id}` response includes `config.interactive` (it already does via `AgentConfig` serialisation — verify this is present in the response body).

### Frontend — `AgentTerminal.tsx`

Add an `agentInteractive` prop (or derive it from an `agent` prop):

```tsx
interface AgentTerminalProps {
  agentId: string
  agentInteractive?: boolean  // true = PTY stdin; false = SDK WebSocket (default false)
  readOnly?: boolean
}
```

When `agentInteractive=false` (SDK mode) and the user enables Interactive mode:
- Input should be sent via `POST /agents/{id}/message` (or a dedicated WebSocket) rather than PTY binary frames
- The toolbar label could read **"Send message"** instead of **"Interactive"** to set correct expectations

When `agentInteractive=true` (interactive mode):
- Current behaviour is correct: binary frames → PTY stdin

### UX note

Consider disabling the Interactive toggle entirely for SDK-mode agents and replacing it with a text input that calls `POST /agents/{id}/message` — this is a clearer affordance than repurposing the existing toggle.

## Acceptance criteria

- [ ] `agentInteractive` prop added to `AgentTerminal`
- [ ] For SDK-mode agents, user input goes via `POST /agents/{id}/message`, not PTY stdin
- [ ] For interactive-mode agents, current PTY stdin path is preserved
- [ ] Tests updated to cover both routing paths
- [ ] Agent detail page passes `agent.config.interactive` to `AgentTerminal`

geoffjay:
fix(ui): route terminal input correctly based on agent mode

fix(ui): route AgentTerminal input based on agent mode

SDK-mode agents (agentInteractive=false, the default) had their keyboard
input sent to PTY stdin via binary WebSocket frames — which Claude ignores
because it reads from its SDK WebSocket, not PTY stdin.

Changes:
- Add agentInteractive prop to AgentTerminal (default false = SDK mode)
- SDK mode: terminal is always read-only; the Interactive toggle is
  replaced by a compact compose input in the toolbar that calls
  POST /agents/{id}/message via orchestratorClient.sendMessage()
- Interactive mode (agentInteractive=true): existing PTY stdin toggle
  behaviour is preserved unchanged
- agentInteractiveRef gates the onData PTY handler as a belt-and-
  suspenders guard (disableStdin=true already prevents onData firing)
- AgentDetail now passes agentInteractive={agent.config.interactive}

Tests (29 total, 9 new):
- Interactive toggle tests scoped to agentInteractive=true
- SDK-mode describe block: compose input visibility, send routing,
  clear-on-success, error banner, disabled states

Closes #773

### DIFF
--- a/ui/src/components/agents/AgentTerminal.tsx
+++ b/ui/src/components/agents/AgentTerminal.tsx
@@ -5,7 +5,12 @@
  * - Full ANSI color/cursor support via xterm.js
  * - Binary WebSocket frames for PTY output; JSON text frames for resize/control
  * - Auto-resize to container via FitAddon + ResizeObserver
- * - Read-only mode (default) with opt-in interactive mode toggle
+ * - Mode-aware input routing:
+ *   - Interactive-mode agents (agentInteractive=true): read-only / interactive
+ *     toggle forwards keystrokes directly to PTY stdin as binary frames.
+ *   - SDK-mode agents (agentInteractive=false, default): terminal is always
+ *     read-only; a compact compose input in the toolbar sends messages via
+ *     POST /agents/{id}/message so they reach Claude over the SDK WebSocket.
  * - Exponential backoff reconnection (matches WebSocketManager behaviour)
  * - Graceful fallback when the backend does not support PTY streaming
  * - Web links addon (clickable URLs in output)
@@ -26,12 +31,14 @@ import {
   KeyboardOff,
   Loader2,
   Search,
+  Send,
   TerminalSquare,
   Wifi,
   WifiOff,
   X,
 } from 'lucide-react'
 import { serviceConfig } from '@/services/config'
+import { orchestratorClient } from '@/services/orchestrator'
 
 // ---------------------------------------------------------------------------
 // Terminal WebSocket URL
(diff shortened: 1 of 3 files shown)


geoffjay:

- #777 ◀

PR #799, opened 2026-03-25
feat: add Linear webhook payload parsing

geoffjay:
feat: add Linear webhook payload parsing

feat(webhook): add Linear webhook payload parsing

- Add `parse_linear_event()` handling Issue, Comment, IssueLabel events
- Update `parse_webhook_payload` signature with `linear_event` and
  `linear_delivery` params; Linear takes dispatch priority over GitHub
- Update `handle_webhook` in api.rs to extract Linear-Event,
  Linear-Delivery, and Linear-Signature headers; choose correct
  signature header based on event source
- Add GitHub and Linear webhook variables to KNOWN_VARIABLES in template.rs
- Add 8 new Linear webhook parsing tests and 1 signature format test

### DIFF
--- a/crates/orchestrator/src/scheduler/api.rs
+++ b/crates/orchestrator/src/scheduler/api.rs
@@ -1,7 +1,7 @@
 use crate::api::ApiError;
 use crate::manager::AgentManager;
 use crate::scheduler::template::validate_template;
-use crate::scheduler::types::*;
+use crate::scheduler::types::{WebhookSource, *};
 use crate::scheduler::webhook;
 use crate::scheduler::Scheduler;
 use crate::types::{clamp_limit, AgentStatus, PaginatedResponse};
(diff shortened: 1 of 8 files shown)


crates/orchestrator/src/scheduler/api.rs
```diff
@@ -392,26 +392,42 @@ async fn handle_webhook(
...
+    let linear_delivery = headers.get("linear-delivery").and_then(|v| v.to_str().ok());
+
     // Verify HMAC-SHA256 signature if a secret is configured.
+    // Linear sends the signature in `Linear-Signature` (raw hex); GitHub uses
+    // `X-Hub-Signature-256` (prefixed with `sha256=`). Both formats are handled
+    // by `verify_signature` which strips the prefix when present.
     if let Some(ref secret_value) = secret {
-        let signature =
-            headers.get("x-hub-signature-256").and_then(|v| v.to_str().ok()).unwrap_or("");
+        let signature = if linear_event.is_some() {
```
geoffjay: [SECURITY — CRITICAL] The `if linear_event.is_some()` branch to select which signature header to verify can be forced by any HTTP caller injecting a `Linear-Event` header into a request targeting a GitHub-configured workflow.

If the workflow has no configured secret, verification is skipped entirely and the fabricated `Linear-Event` alone is enough to force Linear dispatch. If a secret is configured, the HMAC check runs but against the attacker-selected `Linear-Signature` header value.

Recommended fix: store an `expected_source` field in `WebhookEntry` at registration time and reject requests whose source headers do not match before performing signature selection.

PR #866, opened 2026-03-28
epic: FileWatch trigger — filesystem event-driven workflows (#804)

geoffjay:
Rollup PR for epic #804.

This branch collects all sub-issue work via stacked PRs.

Sub-issues will be linked as they are implemented.

PR #867, opened 2026-03-28
feat(scheduler): implement FileWatchStrategy with notify crate

### LINKED ISSUE #810: Implement FileWatchStrategy with notify crate for native filesystem events
## Summary

Implement the core `FileWatchStrategy` using the `notify` crate (v6+) for native OS filesystem event monitoring. This is the primary implementation path providing low-latency file change detection.

## Implementation Details

- Add `notify` (v6+) and `notify-debouncer-full` as dependencies to orchestrator Cargo.toml
- Add `globset` for glob pattern matching on file paths
- Create `FileWatchStrategy` implementing `TriggerStrategy`
- Use `notify::RecommendedWatcher` for cross-platform support (inotify on Linux, FSEvents on macOS, ReadDirectoryChanges on Windows)
- Apply debouncing via `notify-debouncer-full` with configurable timeout (default 500ms)
- Filter events through glob patterns before producing tasks
- Produce `Task` with:
  - `source_id: "file:{relative_path}:{event_kind}"` for deduplication
  - `title`: changed file path
  - `body`: JSON with event details (path, event type, timestamp)
- Support event type filtering: Create, Modify, Remove, Rename
- Use `tokio::sync::mpsc` to bridge the synchronous notify callback to async strategy

## Acceptance Criteria

- [ ] `FileWatchStrategy` implements `TriggerStrategy` trait
- [ ] Uses `notify::RecommendedWatcher` for native OS events
- [ ] Debouncing via `notify-debouncer-full` with configurable duration
- [ ] Glob pattern matching filters which files trigger events
- [ ] Event type filtering (create/modify/remove/rename)
- [ ] Async bridge from notify's sync callbacks to tokio
- [ ] Respects shutdown signal
- [ ] Unit tests with temporary directories and file operations

## Stack Base

Parallel: no ordering constraint with other Phase 6 epics

## Key Files

- `crates/orchestrator/Cargo.toml` — add notify, notify-debouncer-full, globset
- `crates/orchestrator/src/scheduler/strategy.rs` — add `FileWatchStrategy`

geoffjay:
Add FileWatchStrategy that subscribes to OS-native filesystem events (inotify/FSEvents/kqueue) via the `notify` crate, with globset pattern filtering, event-kind filtering (create/modify/delete), and a debounce window that resets on each burst event.

Closes #810

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -435,7 +435,7 @@ dependencies = [
  "metrics",
  "metrics-exporter-prometheus",
  "mockito",
- "notify",
+ "notify 0.2.0",
  "reqwest",
  "serde",
  "serde_json",
@@ -853,6 +853,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63044e1ae8e69f3b5a92c736ca6269b8d12fa7efe39bf34ddb06d102cf0e2cab"
+dependencies = [
+ "memchr",
+ "serde",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1027,7 +1037,7 @@ dependencies = [
  "memory",
  "mockito",
  "monitor",
- "notify",
+ "notify 0.2.0",
  "orchestrator",
  "prettytable-rs",
  "reqwest",
@@ -1309,7 +1319,7 @@ dependencies = [
  "bitflags 2.11.0",
  "crossterm_winapi",
  "futures-core",
- "mio",
+ "mio 1.1.1",
  "parking_lot",
  "rustix 0.38.44",
  "signal-hook",
@@ -2387,6 +2397,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2479,6 +2500,15 @@ checksum = "f7e180ac76c23b45e767bd7ae9579bc0bb458618c4bc71835926e098e61d15f8"
 dependencies = [
  "rustix 0.38.44",
  "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


geoffjay (commented):
## Review: feat(scheduler): implement FileWatchStrategy with notify crate

Well-structured base implementation. The sync-to-async bridge, glob filtering, event-kind filtering, and resettable debounce are all correct. A couple of minor observations, none blocking.

---

### Stack context

`issue-810` is the bottom of a four-deep stack: `epic-804 (#866)` → `issue-810 (#867)` → `issue-811 (#868)` → `issue-812 (#875)` → `issue-813 (#876)`. The `epic-804` base shows `needs restack` against `main`, so the whole stack inherits that status. Adding `needs-restack`.

---

### Dependency choices — correct

`notify = "6"` in `crates/orchestrator/Cargo.toml` is the filesystem watcher crate from crates.io — not `{ path = "../notify" }` which would pull in the agentd notification service. The Cargo.lock correctly shows both `notify 0.2.0` (agentd notification service) and `notify 6.1.1` (filesystem watcher) as distinct packages. ✅

`globset = "0.4"` pinned directly, consistent with other non-workspace deps in this crate. ✅

---

### Sync-to-async bridge — correct

```rust
let (tx, rx) = mpsc::unbounded_channel();
let mut watcher = notify::RecommendedWatcher::new(
    move |res: notify::Result<notify::Event>| {
        let _ = tx.send(res);
    },
    notify::Config::default(),
)?;
```

`notify`'s watcher callback runs on a background thread (sync context). Bridging through `mpsc::unbounded_channel` is the correct pattern — the `UnboundedSender` is `Send`, and `let _ = tx.send(res)` ignores errors that occur when the receiver has been dropped during shutdown. ✅

`_watcher` held in the struct as a drop guard is correct — dropping the watcher stops watching and the background thread exits. ✅

---

### `next_tasks` — correct two-phase logic

**Phase 1 (first event):** The `select!` loop correctly handles all four cases: matching event (break), non-matching event (continue), watcher error (log + continue), channel closed (return empty). The `shutdown.changed()` arm checks `*shutdown.borrow()` after firing, which is the right guard against spurious wakeups if the value is still `false`. ✅

**Phase 2 (debounce):** The resettable timer pattern is correct:
```rust
loop {
    let sleep = tokio::time::sleep(debounce);
    tokio::pin!(sleep);
    tokio::select! {
        _ = &mut sleep => { break; }        // window expired
        maybe_event = self.rx.recv() => { /* loop to reset */ }
        ...
    }
}
```
A new `sleep` future is created at the top of each iteration. When any event arrives, the loop restarts with a fresh `debounce`-duration sleep, effectively resetting the timer. The old pinned future is dropped. This is idiomatic tokio and correct. ✅

---

### Minor: debounce resets on non-matching events

During the debounce phase, `Some(Ok(_))` matches all received events — matching or not. A non-matching event (wrong kind, excluded by glob) still resets the debounce timer. This is a conservative design choice (any filesystem activity in the watched tree extends the quiet period), but in high-noise directories (e.g., watching `/var/log` for a specific pattern) the debounce window may never expire. Worth a doc comment, not a blocker.

---

### Minor: `"rename"` in event_kinds doc comment is a dead filter value

The doc comment lists `"create"`, `"modify"`, `"delete"`, `"access"`, `"rename"` as accepted event kind strings. However, `event_kind_to_str` never returns `"rename"` — rename events in `notify` arrive as `Modify(ModifyKind::Name(_))` and map to `"modify"`. A caller filtering on `"rename"` would never receive a task. This inaccuracy is already corrected by PR #876 at the top of the stack (which removes `"rename"` from the doc). Non-blocking since it's a doc-only issue fixed upstack.

---

### `build_task` — correct

All five required metadata keys are populated: `file_path`, `file_name`, `file_dir`, `event_type`, `timestamp`. The `source_id` format `"file:<path>:<kind>"` is consistent with how the template variable `event_type` references these in PR #875. ✅

Using `event.paths.first()` for single-file events is correct; multi-path events (rare, mostly internal notify bookkeeping) silently use only the first path. ✅

---

### `matches_event` — correct

Empty `event_kinds` → accept all; patterns `None` → accept all. The `any(|k| k == kind_str)` and `any(|p| patterns.is_match(p))` semantics are correct (OR across patterns, not AND). ✅

---

### Glob ordering note

In `new()`, the glob set is built after the watcher is created and paths are registered. If glob validation fails, the native watcher is dropped immediately (no background task to clean up), so there's no resource leak. The "validate before spawn" concern is more critical for the polling task in PR #868. For this native-only PR the ordering is acceptable — but the fix in PR #876 (validating before spawning) is the right long-term approach for consistency. ✅

---

### Summary

Correct, clean base implementation. The two minor observations (debounce-on-all-events, "rename" doc inaccuracy) are both addressed further up the stack. Approving; restacking needed once `epic-804` is current with main.

PR #930, opened 2026-04-02
fix(ui): add memory and communicate service proxy routes

### LINKED ISSUE #929: fix(ui): Add memory and communicate service proxy routes to UI server
## Context

The UI server (`crates/ui/`) acts as a reverse proxy for backend services, but only proxies three of the five services:

| Service | Proxy Route | Status |
|---------|------------|--------|
| ask | `/api/ask/**` | Proxied |
| notify | `/api/notify/**` | Proxied |
| orchestrator | `/api/orchestrator/**` | Proxied |
| **memory** | -- | **Missing** |
| **communicate** | -- | **Missing** |

Because the memory service is not proxied, the React frontend calls it directly from the browser using `serviceConfig.memoryServiceUrl`, which defaults to `http://localhost:17008` (dev port). However, the memory service is running on production port **7008**. This causes the `/memories` page to silently fail every fetch -- the page appears empty even though memories exist and are accessible via the CLI.

The communicate service has the same gap but may be less visible since it also uses WebSocket connections.

### How to reproduce

1. Start all services (memory on port 7008, UI on port 7009)
2. Create a memory via CLI: `agent memory remember "test" --created-by planner --type information`
3. Verify via CLI: `agent memory search "test" --as-actor planner` -- works fine
4. Open the UI at `http://localhost:7009/memories` -- shows no memories

### Root cause

- `crates/ui/src/lib.rs` only registers proxy routes for ask, notify, orchestrator
- `crates/ui/src/proxy.rs` only has `ProxyState` fields and handlers for those three
- `crates/ui/src/config.rs` only loads env vars for those three service URLs
- `ui/src/services/config.ts` has the memory URL but the frontend hits it directly from the browser, which fails due to port mismatch (17008 vs 7008)

### 1. Add memory and communicate to `ProxyState` (`crates/ui/src/proxy.rs`)

```rust
pub struct ProxyState {
    pub client: Client,
    pub ask_url: String,
    pub notify_url: String,
    pub orchestrator_url: String,
    pub memory_url: String,        // new
    pub communicate_url: String,   // new
}
```

Add proxy handlers:

```rust
pub async fn proxy_memory(...) -> Result<Response, StatusCode> {
    proxy_request(&state.client, &state.memory_url, "/api/memory", req).await
}

pub async fn proxy_communicate(...) -> Result<Response, StatusCode> {
    proxy_request(&state.client, &state.communicate_url, "/api/communicate", req).await
}
```

### 2. Add config entries (`crates/ui/src/config.rs`)

```rust
pub memory_service_url: String,
pub communicate_service_url: String,
```

With env vars `AGENTD_MEMORY_SERVICE_URL` (default `http://localhost:7008`) and `AGENTD_COMMUNICATE_SERVICE_URL` (default `http://localhost:7010`).

### 3. Register routes (`crates/ui/src/lib.rs`)

```rust
.route("/api/memory/{*path}", any(proxy::proxy_memory))
.route("/api/communicate/{*path}", any(proxy::proxy_communicate))
```

### 4. Update React frontend to use proxy (`ui/src/services/config.ts`)

Change the frontend to route through the UI proxy instead of calling services directly:

```typescript
export const serviceConfig = {
    askServiceUrl: "/api/ask",
    notifyServiceUrl: "/api/notify",
    orchestratorServiceUrl: "/api/orchestrator",
    memoryServiceUrl: "/api/memory",
    communicateServiceUrl: "/api/communicate",
} as const;
```

This eliminates the port mismatch entirely -- the browser only talks to the UI server, which proxies to the correct backend ports.

## Acceptance Criteria

- [ ] Memory service proxied through UI at `/api/memory/**`
- [ ] Communicate service proxied through UI at `/api/communicate/**`
- [ ] `UiConfig` loads `AGENTD_MEMORY_SERVICE_URL` and `AGENTD_COMMUNICATE_SERVICE_URL` env vars
- [ ] React frontend uses relative proxy paths instead of direct service URLs
- [ ] `/memories` page shows memories created via CLI
- [ ] No CORS or port mismatch issues when accessing any service through the UI

## Relevant Files

- `crates/ui/src/lib.rs` -- router setup (add routes)
- `crates/ui/src/proxy.rs` -- ProxyState and handlers (add memory + communicate)
- `crates/ui/src/config.rs` -- UiConfig (add service URLs)
- `ui/src/services/config.ts` -- frontend service URLs (switch to proxy paths)

## Stack Base
Parallel: no ordering constraint

geoffjay:
Add proxy routes for the memory and communicate services to the UI server, and update the React frontend to use relative proxy paths instead of direct localhost URLs. This fixes the /memories page silently failing due to port mismatch (17008 vs 7008).

Closes #929

### DIFF
--- a/crates/ui/src/config.rs
+++ b/crates/ui/src/config.rs
@@ -10,6 +10,10 @@ pub struct UiConfig {
     pub notify_service_url: String,
     /// URL of the orchestrator service.
     pub orchestrator_service_url: String,
+    /// URL of the memory service.
+    pub memory_service_url: String,
+    /// URL of the communicate service.
+    pub communicate_service_url: String,
 }
 
 impl UiConfig {
@@ -20,6 +24,8 @@ impl UiConfig {
     /// - `AGENTD_ASK_SERVICE_URL` — ask service URL (default: `http://localhost:7001`)
     /// - `AGENTD_NOTIFY_SERVICE_URL` — notify service URL (default: `http://localhost:7004`)
     /// - `AGENTD_ORCHESTRATOR_SERVICE_URL` — orchestrator service URL (default: `http://localhost:7006`)
+    /// - `AGENTD_MEMORY_SERVICE_URL` — memory service URL (default: `http://localhost:7008`)
+    /// - `AGENTD_COMMUNICATE_SERVICE_URL` — communicate service URL (default: `http://localhost:7010`)
     pub fn from_env() -> Self {
         Self {
             port: std::env::var("AGENTD_PORT").ok().and_then(|v| v.parse().ok()).unwrap_or(17009),
@@ -30,6 +36,10 @@ impl UiConfig {
                 .unwrap_or_else(|_| "http://localhost:7004".to_string()),
             orchestrator_service_url: std::env::var("AGENTD_ORCHESTRATOR_SERVICE_URL")
                 .unwrap_or_else(|_| "http://localhost:7006".to_string()),
+            memory_service_url: std::env::var("AGENTD_MEMORY_SERVICE_URL")
+                .unwrap_or_else(|_| "http://localhost:7008".to_string()),
+            communicate_service_url: std::env::var("AGENTD_COMMUNICATE_SERVICE_URL")
+                .unwrap_or_else(|_| "http://localhost:7010".to_string()),
         }
     }
 }
--- a/crates/ui/src/lib.rs
+++ b/crates/ui/src/lib.rs
@@ -22,6 +22,8 @@
 //! - `AGENTD_ASK_SERVICE_URL` — Ask service URL (default: `http://localhost:7001`)
 //! - `AGENTD_NOTIFY_SERVICE_URL` — Notify service URL (default: `http://localhost:7004`)
 //! - `AGENTD_ORCHESTRATOR_SERVICE_URL` — Orchestrator service URL (default: `http://localhost:7006`)
+//! - `AGENTD_MEMORY_SERVICE_URL` — Memory service URL (default: `http://localhost:7008`)
+//! - `AGENTD_COMMUNICATE_SERVICE_URL` — Communicate service URL (default: `http://localhost:7010`)
 //! - `RUST_LOG` — Logging level (default: info)
 
 pub mod config;
@@ -53,6 +55,8 @@ pub async fn run(config: config::UiConfig) -> Result<()> {
         ask_url: config.ask_service_url,
         notify_url: config.notify_service_url,
         orchestrator_url: config.orchestrator_service_url,
+        memory_url: config.memory_service_url,
+        communicate_url: config.communicate_service_url,
     };
 
     // SPA fallback: serve index.html for any path that doesn't match a file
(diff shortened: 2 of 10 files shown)


geoffjay (commented):
## Review: fix(ui): add memory and communicate service proxy routes

The core goal is correct and the implementation approach is sound, but there are two issues to address before merge.

---

### Blocker 1: Branch needs restack

The diff context reveals this branch was created before the `index_url` proxy support landed on `main`. The current `main` branch has `index_url: String` in `ProxyState` (proxy.rs), `index_service_url: String` in `UiConfig` (config.rs), and the matching `proxy_index` route wired in `lib.rs` — none of which appear in this PR's diff context. Merging as-is would produce a conflict or silently drop the index service proxy.

**Fix:** run `git-spice upstack restack` to pull in the index service changes, then re-push.

---

### Blocker 2: Rust proxy does not support WebSocket upgrades

`config.ts` is changed to route **all** service URLs through the proxy — including the orchestrator and communicate services, both of which have WebSocket endpoints:

- `AgentTerminal.tsx` opens `ws://<host>/api/orchestrator/terminal/<id>`
- `useAgentStream.ts` / `useAllAgentsStream.ts` open `ws://<host>/api/orchestrator/stream/<id>`
- `useCommunicateSocket.ts` opens `ws://<host>/api/communicate/...`

The Rust proxy in `proxy_request` uses `reqwest::Client`, which is an HTTP client and does not implement the WebSocket upgrade handshake (HTTP 101 / `Upgrade: websocket`). When the browser initiates a WebSocket connection through `/api/orchestrator` or `/api/communicate`, the proxy will forward the upgrade request to the upstream but then call `.bytes().await` on the response — which will either hang waiting for the stream to complete or return a malformed response, breaking the WebSocket connection.

The Vite dev proxy handles this correctly with `ws: true`, but that only applies in local development. The production Rust proxy has no equivalent.

**Before this PR:** WebSocket connections used direct `http://localhost:17006` URLs (baked in at build time), bypassing the proxy entirely. **After this PR:** they are routed through the Rust proxy and will fail.

**Options to fix:**

1. **Lowest-risk / recommended for this PR's scope:** Keep WebSocket-bearing services on absolute URLs and only route REST-only services (memory, ask, notify) through the proxy. You could split the orchestrator config into separate REST and WS URL keys, or leave orchestrator/communicate on direct URLs for now.
2. Replace `reqwest`-based proxying with a crate that supports HTTP upgrade tunneling (e.g. `hyper-reverse-proxy` with upgrade support).
3. Add dedicated WebSocket tunnel handlers alongside the HTTP proxy handlers for services that need it.

Option 1 unblocks the memories fix without regressing agent streaming or communicate.

---

### Minor: `ws: true` on memory proxy is unnecessary

```ts
"/api/memory": {
    target: memoryServiceUrl,
    ws: true,   // memory is REST-only — this flag is misleading
```

The memory service appears to be HTTP/REST only. `ws: true` is harmless but misleading. Worth removing for clarity. (Separately, the existing `/api/orchestrator` Vite entry is missing `ws: true`, which is the actual gap for agent streaming in dev — but that's a pre-existing issue outside this PR's scope.)

---

### What's good

- The Rust-side additions (`config.rs`, `lib.rs`, `proxy.rs`) are clean and follow the existing proxy pattern exactly — naming, error handling, doc comments are all consistent.
- The `base.ts` relative-path guard (`startsWith("/") ? window.location.origin + ...`) is correct and handles both absolute and relative base URLs properly.
- The same guard is applied consistently across all four hook/component files that construct WebSocket URLs.
- The vite dev proxy entries for the new services are structured correctly, and `ws: true` on communicate is appropriate for that service.
- The PR description accurately describes the problem (17008 vs 7008 port mismatch) and the intended solution.

PR #1129, opened 2026-04-16
feat(auth): gateway completeness, TenantId extractor, POST /auth/validate, Procfile

### LINKED ISSUE #1116: Auth integration foundation: gateway completeness, tenant extractor, token validation
## Context

The `agentd-core` service has a complete auth system and a gateway proxy that injects `X-Tenant-ID` headers to downstream services, but nothing uses it yet. This issue establishes the infrastructure that all subsequent auth integration work depends on.

**Architecture**: Gateway-centric. All external traffic (CLI, UI) routes through the core gateway at `/api/v1/{service}/*`. Core validates the bearer token, resolves the active organization, and injects `X-Tenant-ID` + `X-Request-ID` headers. Downstream services read `X-Tenant-ID` via a shared extractor from `agentd-common`.

**Exceptions**: WebSocket endpoints validate tokens directly at upgrade time. MCP server stays local/trusted with no auth.

## Checklist

- [ ] **Add missing services to core gateway proxy** -- The gateway in `crates/core/src/proxy.rs` only covers 6 of 9 services. Add `memory` (MEMORY_URL, default `http://localhost:17008`), `communicate` (COMMUNICATE_URL, default `http://localhost:17010`), and `index` (INDEX_URL, default `http://localhost:17012`) to `ProxyConfig::from_env()`. Update `test_proxy_config_defaults`.
  - Files: `crates/core/src/proxy.rs`

- [ ] **Create X-Tenant-ID extractor in agentd-common** -- Add a `tenant` module with two Axum extractors: `TenantId` (reads `X-Tenant-ID` header, returns 400 if missing) and `OptionalTenantId` (returns `Option<String>`, never rejects). Controlled by `AGENTD_REQUIRE_TENANT` env var (default `false`). When false, `TenantId` falls back to permissive mode so services work without the gateway in dev/test.
  - Files: `crates/common/src/tenant.rs` (new), `crates/common/src/lib.rs`

- [ ] **Add token validation endpoint to core service** -- Add `POST /auth/validate` that accepts a bearer token and returns `{ user_id, organization_id, valid: true }`. Used by WebSocket endpoints to validate tokens at connection time. Should be fast (single DB lookup).
  - Files: `crates/core/src/api/auth.rs`

- [ ] **Add core service to Procfile** -- Current Procfile is missing core, wrap, communicate, memory, and index.
  - Files: `Procfile`

## Dependencies

None. This is the first phase -- all other auth integration work depends on it.

geoffjay:
Implements all four checklist items from the auth integration foundation issue:

- **Gateway completeness**: adds `memory` (17008), `communicate` (17010), and `index` (17012) to `ProxyConfig::from_env()` with env-override support and expanded tests
- **TenantId extractors**: new `agentd-common::tenant` module with `TenantId` (permissive/strict via `AGENTD_REQUIRE_TENANT`) and `OptionalTenantId` (always succeeds, returns `Option<String>`) — 5 tests covering all modes
- **Token validation endpoint**: `POST /auth/validate` in core service returns `{ valid, user_id, organization_id }` for fast WebSocket token checks at upgrade time — 3 tests
- **Procfile**: adds `core`, `wrap`, `communicate`, `memory`, `index` entries

Closes #1116


geoffjay:

- #1135
    - #1129 ◀

geoffjay (commented):
## Review: feat(auth): gateway completeness, TenantId extractor, POST /auth/validate, Procfile

**LGTM** — cannot self-approve, posting as comment.

### Stack note

This branch (issue-1116) targets `feature/core`, which itself targets `main`. It is the direct parent of the entire core migration stack (#1130 → #1131 → #1132 → #1133 / #1134). Those PRs can stack and merge after this one lands.

### What was verified

**Procfile** — 5 new entries; binary names (`communicate`, `agentd-core`, `agentd-index`, `memory`, `wrap`) match the `package.name` convention already used in the file. Alphabetical ordering is preserved across all entries. ✓

**`crates/common/src/tenant.rs`** — two Axum extractors:
- `TenantId`: header present → wraps value; absent + strict → `MissingTenantId` (400 JSON); absent + permissive → `TenantId(String::new())`. Logic is correct; `MissingTenantId` uses `IntoResponse` cleanly. ✓
- `OptionalTenantId`: `Rejection = Infallible`, always succeeds, returns `Option<String>`. ✓
- Tests: both permissive/strict modes combined in one function to avoid env-var races across parallel test threads — same defensive pattern used in `test_get_db_path_respects_agentd_env` in `storage.rs`. ✓

**`POST /auth/validate`** — delegates entirely to the existing `AuthUser` extractor for auth enforcement, so the 401 path requires no new logic (extractor rejects before the handler body runs). The 200 path is a pure read of `auth.user` fields — no session mutation, no DB write. ✓
- `ValidateResponse.valid` is always `true` in the 200 body (can never be `false`; invalid tokens produce 401 before reaching the handler). Self-documenting for WebSocket upgrade callers. Acceptable.
- 3 tests cover: valid token → 200, invalid token → 401, missing token → 401. ✓

**`crates/core/src/proxy.rs`** — 3 new entries with port assignments:
- `memory` → 17008, `communicate` → 17010, `index` → 17012
- No collisions with existing assignments (17001–17006). Module doc table updated to match. ✓
- `test_proxy_config_new_service_defaults` pins the default values; `test_proxy_config_all_services` extended to include new names. ✓

### Non-blocking suggestions

1. **`tenant_required()` reads the env var on every request.** For a hot extractor path this is a small but repeated syscall. A `std::sync::OnceLock` initialised at first call would be faster and also remove the surprising runtime-change behaviour. Since this is consistent with how `AGENTD_ENV` is read in `get_db_path`, no urgency — but worth a follow-up when the tenant module gets wider adoption.

2. **`TenantId(String::new())` in permissive mode.** Documented, intentional — but service authors using `tenant_id.0` as a DB scope key will silently filter by empty string rather than returning all rows or failing. A one-line `// NOTE: empty string means no gateway in dev; callers should treat this as "no tenant scope".` comment on the `None => Ok(TenantId(String::new())))` arm would save future confusion.

PR #1131, opened 2026-04-16
feat(core): expose apply/status/rollback migration helpers from lib.rs

geoffjay:
Adds `apply_migrations_for_path`, `migration_status_for_path`, and `rollback_migrations_for_path` to `agentd_core` lib.rs, following the pattern established by agentd-orchestrator. Required by the migrate CLI subcommand.

Closes #1124

PR #1130, opened 2026-04-16
feat(common): add rollback_migrations helper to agentd_common::storage

geoffjay:
Adds `rollback_migrations<M>(db_path, steps)` to `agentd_common::storage`, following the same pattern as `apply_migrations` and `migration_status`. `steps=None` rolls back all applied migrations; `steps=Some(n)` rolls back the n most recent.

Closes #1123

PR #1132, opened 2026-04-16
feat(core): add clap CLI with migrate subcommand to agentd-core binary

### LINKED ISSUE #1125: Add clap CLI with migrate subcommand to agentd-core binary
## Context

The `agentd-core` binary currently has no CLI argument parsing - it immediately starts the HTTP server. Issue #1065 requires a `migrate` subcommand for operators to inspect and manage migration state without starting the full service.

## Implementation

Restructure `crates/core/src/main.rs` to use clap with two top-level subcommands:

### CLI Structure

```
agentd-core serve              # Start the HTTP server (current behavior, also the default)
agentd-core migrate status     # List migrations and applied/pending state
agentd-core migrate up         # Apply all pending migrations
agentd-core migrate down       # Roll back the most recent migration (requires --yes)
agentd-core migrate down --all # Roll back all migrations (requires --yes)
```

Running `agentd-core` with no arguments should default to `serve` to maintain backward compatibility.

### Key implementation details

1. **Add `clap` dependency** to `crates/core/Cargo.toml`
2. **Define CLI structs** using clap derive macros:
   ```rust
   #[derive(Parser)]
   #[command(name = "agentd-core")]
   struct Cli {
       #[command(subcommand)]
       command: Option<Command>,
   }

   #[derive(Subcommand)]
   enum Command {
       /// Start the HTTP server (default)
       Serve,
       /// Manage database migrations
       Migrate {
           #[command(subcommand)]
           action: MigrateAction,
       },
   }

   #[derive(Subcommand)]
   enum MigrateAction {
       /// Show migration status
       Status,
       /// Apply all pending migrations
       Up,
       /// Roll back migrations
       Down {
           /// Roll back all migrations instead of just the latest
           #[arg(long)]
           all: bool,
           /// Skip confirmation prompt
           #[arg(long)]
           yes: bool,
       },
   }
   ```
3. **Confirmation for `down`**: When `--yes` is not passed, print which migration(s) will be rolled back and prompt for confirmation on stdin. If stdin is not a TTY, exit with an error requesting `--yes`.
4. **Output formatting**: Use `colored` crate (already a transitive dependency) for status/up/down output, matching the style used in xtask's migrate commands.
5. **Database path**: Use `agentd_common::storage::get_db_path("agentd-core", "core.db")` - same path the server uses, respecting `AGENTD_ENV`.
6. **Exit codes**: Exit 0 on success, non-zero on error with a clear message via anyhow.

### Example output

```
$ agentd-core migrate status
Migration Status (agentd-core):
  ✓ applied  m20260305_000001_create_core_tables
  ✓ applied  m20260408_000002_add_username_to_users
  2 applied, 0 pending

$ agentd-core migrate up
Applying migrations...
  ✓ up to date (2 applied, 0 pending)

$ agentd-core migrate down --yes
Rolling back latest migration...
  ✓ rolled back m20260408_000002_add_username_to_users
```

## Acceptance Criteria

- [ ] `agentd-core` with no args starts the HTTP server (backward compatible)
- [ ] `agentd-core serve` starts the HTTP server
- [ ] `agentd-core migrate status` prints each migration name and applied/pending state
- [ ] `agentd-core migrate up` applies all pending migrations and exits 0
- [ ] `agentd-core migrate down` requires `--yes` flag or interactive confirmation
- [ ] `agentd-core migrate down --all --yes` rolls back all migrations
- [ ] Uses `AGENTD_ENV`-aware database path via `get_db_path`
- [ ] Exits non-zero on error with a clear message
- [ ] `clap` added to `Cargo.toml` dependencies

## Relevant Files

- `crates/core/src/main.rs` - restructure with clap CLI
- `crates/core/Cargo.toml` - add `clap` dependency
- `crates/core/src/lib.rs` - migration helper functions (from #1124)

## Blocked By
- #1124 (migration convenience helpers must be exposed from core lib.rs)

## Stack Base
Branch off `feature/autonomous-pipeline`.
Blocked by: #1124

geoffjay:
Restructures `main.rs` to use clap with `serve` (default) and `migrate status|up|down` subcommands. Adds `--db-path` override for isolated testing, TTY/`--yes` guard on `down`, and human-friendly output. Adds `clap 4` and `atty 0.2` dependencies.

Closes #1125

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,6 +43,7 @@ dependencies = [
  "async-trait",
  "axum",
  "chrono",
+ "clap",
  "http-body-util",
  "hyper",
  "metrics",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -32,8 +32,7 @@ async-trait = "0.1"
 argon2 = "0.5"
 rand = "0.8"
 reqwest = { workspace = true, features = ["json", "stream"] }
-clap = { version = "4", features = ["derive"] }
-atty = "0.2"
+clap = { workspace = true }
 
 [dev-dependencies]
 tower = { version = "0.5", features = ["util"] }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -32,6 +32,8 @@ async-trait = "0.1"
 argon2 = "0.5"
 rand = "0.8"
 reqwest = { workspace = true, features = ["json", "stream"] }
+clap = { version = "4", features = ["derive"] }
+atty = "0.2"
 
 [dev-dependencies]
 tower = { version = "0.5", features = ["util"] }
--- a/crates/core/src/main.rs
+++ b/crates/core/src/main.rs
@@ -22,6 +22,7 @@
 //! | `AGENTD_LOG_FORMAT` | (text)  | Set to `json` for JSON |
 //! | `AGENTD_ENV`        | (prod)  | `development`/`test`   |
 
+use std::io::IsTerminal;
 use std::path::PathBuf;
 
 use agentd_common::config::ValidateConfig;
@@ -212,7 +213,7 @@ async fn run_migrate_down(all: bool, yes: bool, db_path_override: Option<PathBuf
 
     // Require explicit confirmation when no TTY is attached.
     if !yes {
-        let is_tty = atty::is(atty::Stream::Stdin);
+        let is_tty = std::io::stdin().is_terminal();
         if !is_tty {
             anyhow::bail!(
                 "stdin is not a TTY — pass --yes to confirm rollback without a prompt.\n\
--- a/crates/core/src/main.rs
+++ b/crates/core/src/main.rs
@@ -2,6 +2,17 @@
 //!
 //! Central authentication and API gateway service for agentd.
 //!
+//! # Subcommands
+//!
+//! ```text
+//! agentd-core                     # default: start HTTP server
+//! agentd-core serve               # start HTTP server (explicit)
+//! agentd-core migrate status      # show migration status
+//! agentd-core migrate up          # apply all pending migrations
+//! agentd-core migrate down        # roll back latest migration (requires --yes or TTY)
+//! agentd-core migrate down --all  # roll back all migrations (requires --yes or TTY)
+//! ```
+//!
 //! # Environment Variables
 //!
 //! | Variable            | Default | Description            |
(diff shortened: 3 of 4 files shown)


geoffjay:

- #1135
    - #1130
        - #1131
            - #1132 ◀
                - #1133

geoffjay (commented):
## Review: feat(core): add clap CLI with migrate subcommand to agentd-core binary

Two blocking issues before this can merge. The overall structure and logic are solid — the subcommand design, TTY guard, output strings, and `--db-path` override are all correct and well-implemented. Cannot self-approve/request-changes; marking `needs-rework` via label.

### Blocking: replace `atty` with `std::io::IsTerminal`

`atty = "0.2"` carries [RUSTSEC-2021-0145](https://rustsec.org/advisories/RUSTSEC-2021-0145) (soundness advisory — unsound use of `from_raw_fd`). It is also unmaintained. More importantly, **the project already uses the stdlib alternative**: `crates/cli/src/picker.rs` imports `std::io::IsTerminal`. Adding a known-unsound crate when the codebase already has the correct pattern is not acceptable.

**Fix** — remove `atty` from `Cargo.toml` entirely, then in `run_migrate_down`:
```rust
// Before
let is_tty = atty::is(atty::Stream::Stdin);

// After
use std::io::IsTerminal;
let is_tty = std::io::stdin().is_terminal();
```
No other changes needed — the call-site logic is correct as written.

### Blocking: use the workspace `clap` instead of a local version spec

The workspace already declares `clap = { version = "4.5", features = ["derive"] }` in `[workspace.dependencies]`. This PR adds its own `clap = { version = "4", features = ["derive"] }` — a different version spec that bypasses the workspace pin.

**Fix** in `crates/core/Cargo.toml`:
```toml
# Before
clap = { version = "4", features = ["derive"] }

# After
clap = { workspace = true }
```
The workspace entry already includes the `derive` feature, so no local feature declaration is needed.

### Non-blocking: `init_tracing()` runs for all subcommands

`agentd_common::server::init_tracing()` is called unconditionally in `main` before dispatch. For `migrate status/up/down` this initialises the full tracing stack when the user likely only wants clean line-oriented output. Won't cause bugs (tracing goes to stderr; integration test stdout assertions still pass), but it's slightly heavy for a CLI utility path. Consider moving the call inside `run_serve()` only. Non-blocking, can be a follow-up.

### Stack note

This branch (issue-1125) is stacked on PR #1131 (issue-1124), which is **LGTM** and can merge first. The integration tests in #1133 depend on this PR's CLI shape (subcommand names, `--db-path` flag, output strings) — those all remain correct and will not need changes once the two blocking issues above are fixed.

geoffjay (commented):
## Review: feat(core): add clap CLI with migrate subcommand to agentd-core binary

### Stack Position

```
issue-1127 (#1133) — test(core): integration tests  [needs push, child]
  └── issue-1125 (#1132) ← this PR
        └── issue-1124 (#1131) — expose migration helpers  [merge-queue ✓]
              └── issue-1123 (#1130)  [merge-queue ✓]
```

Parent #1131 is already in `merge-queue`. Once it lands, this PR is unblocked. Tests for this feature are in child PR #1133 — that split is clean.

---

### API Correctness

I cross-checked each call against `lib.rs` on `issue-1124`:

| Call | Signature | Verdict |
|------|-----------|---------|
| `migration_status_for_path(&db_path)` | `-> Result<Vec<(String, bool)>>` | ✓ |
| `apply_migrations_for_path(&db_path)` | `-> Result<()>` | ✓ |
| `rollback_migrations_for_path(&db_path, None)` | `None` = roll back all | ✓ |
| `rollback_migrations_for_path(&db_path, Some(1u32))` | `Some(n)` = n most-recently-applied | ✓ |

The `steps` computation (`None` for `--all`, `Some(1)` for single step) correctly matches the documented semantics.

---

### TTY / Confirmation Guard

The `run_migrate_down` guard is correctly implemented:

```rust
let is_tty = std::io::stdin().is_terminal();
if !is_tty {
    anyhow::bail!("stdin is not a TTY — pass --yes to confirm rollback without a prompt.\n\
                   Would roll back: {}", targets.join(", "));
}
```

Uses `std::io::IsTerminal` (stable since Rust 1.70) rather than the deprecated `atty` crate — better than what the PR description claims. The error message is actionable: it tells the caller exactly what flag to pass and what would be affected.

---

### One Non-Blocking Note

**PR description inaccuracy:** The description says "Adds `clap 4` and `atty 0.2` dependencies." No `atty` dependency is added (correct!), and `std::io::IsTerminal` is used instead. Worth updating the description to avoid confusion for anyone reading the history, but no code change needed.

**Minor:** `run_serve` calls `agentd_common::storage::get_db_path("agentd-core", "core.db")` directly instead of `resolve_db_path(None)`. Functionally identical since `resolve_db_path(None)` does exactly that, but using the helper would make `run_serve` consistent with the rest of the file. Non-blocking.

---

### What's Working Well

- Clean clap 4 derive structure: `Option<Command>` default → serve, nested `MigrateAction` subcommands, `--db-path` override on every migrate action for isolated testing
- The TTY guard is applied before the destructive call, not after
- `applied.is_empty()` early-return prevents the index operation on an empty vec
- `run_migrate_up` re-queries status after applying to report final counts rather than trusting the pre-run snapshot
- `init_tracing()` called before `Cli::parse()` — correct initialization order
- No `atty` crate: avoids the known soundness issue in that crate

---

### Summary

No blocking issues. The implementation is correct, the destructive-action guard is properly implemented, and the clap structure is clean. Approved pending resolution of parent #1131 (already in merge-queue).

PR #1167, opened 2026-04-17
feat(orchestrator): add conversation_events table, entity, and storage

geoffjay:
Adds the persistence foundation for agent conversation history — the `conversation_events` SQLite table, SeaORM entity, domain types, and four CRUD methods on `AgentStorage`.

## Changes

- **Migration #16** (`m20260417_000016_add_conversation_events.rs`) — creates `conversation_events` table with two composite indexes (`agent_id+created_at`, `agent_id+event_type`); `down()` correctly drops indexes before the table
- **Entity** (`entity/conversation_event.rs`) — SeaORM `Model` with 7 fields matching the schema; `#![allow(dead_code)]` suppresses expected unused-derive warnings until #1160 wires it up
- **Domain types** (`types.rs`) — `ConversationEventType` enum (8 variants, `Display`/`FromStr`), `ConversationEvent` struct with `new()` helper, `ConversationQuery` for filtering/pagination
- **Storage methods** (`storage.rs`) — `insert_conversation_event`, `list_conversation_events` (with type/time/pagination filters), `count_conversation_events`, `delete_conversation_events_for_agent`; plus `model_to_conversation_event` helper

## Test plan

- [x] 7 new unit tests in `storage::tests`: insert+count, list all, filter by type, limit+offset, delete (cross-agent isolation), metadata JSON roundtrip, all 8 event-type variants
- [x] `cargo test -p orchestrator` — 413 tests pass (pre-existing `linear_fetch_http` network tests ignored)
- [x] `cargo fmt --check -p orchestrator` — clean
- [x] `cargo clippy -p orchestrator` — no errors

## Dependencies
- Unblocks: #1160 (WebSocket persistence), #1161 (REST API), #1163 (retention policy)

Closes #1159

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1159, opened 2026-04-17
Add conversation_events table and SeaORM entity to orchestrator


geoffjay:
## Summary

Create a new `conversation_events` SQLite table and corresponding SeaORM entity in the orchestrator crate to durably store all agent conversation events (output, tool use, thinking, results, prompts, status changes).

## Context

The orchestrator's WebSocket handler (`crates/orchestrator/src/websocket.rs`) already receives and parses every message from Claude Code agents. These messages are broadcast to live `/stream` subscribers but never stored. This issue adds the persistence foundation.

## Acceptance Criteria

- [ ] New migration file `m20260417_000014_add_conversation_events.rs` in `crates/orchestrator/src/migration/`
- [ ] Migration registered in `crates/orchestrator/src/migration/mod.rs`
- [ ] New SeaORM entity at `crates/orchestrator/src/entity/conversation_event.rs`
- [ ] Entity registered in `crates/orchestrator/src/entity/mod.rs`
- [ ] New storage methods on `AgentStorage` in `crates/orchestrator/src/storage.rs`:
  - `insert_conversation_event(event: &ConversationEvent) -> Result<Uuid>`
  - `list_conversation_events(agent_id: Uuid, opts: ConversationQuery) -> Result<Vec<ConversationEvent>>`
  - `count_conversation_events(agent_id: Uuid) -> Result<u64>`
  - `delete_conversation_events_for_agent(agent_id: Uuid) -> Result<u64>`
- [ ] New domain types in `crates/orchestrator/src/types.rs`:
  - `ConversationEvent` struct
  - `ConversationEventType` enum (output, tool_use, thinking, result, prompt_sent, activity_changed, usage_update, context_cleared)
  - `ConversationQuery` struct (pagination, type filters, time range)

## Schema Design

```sql
CREATE TABLE conversation_events (
    id TEXT PRIMARY KEY,            -- UUID
    agent_id TEXT NOT NULL,         -- FK to agents.id
    event_type TEXT NOT NULL,       -- 'output', 'tool_use', 'thinking', 'result', 'prompt_sent', 'activity_changed', 'usage_update', 'context_cleared'
    session_number INTEGER NOT NULL DEFAULT 0,  -- increments on context_cleared
    content TEXT,                   -- text content (output lines, thinking text, result text)
    metadata TEXT,                  -- JSON blob for structured data (tool_input, usage stats, etc.)
    created_at TEXT NOT NULL        -- RFC3339 timestamp
);
[...]

PR #1166, opened 2026-04-17
feat(wrap): make PTY ring-buffer configurable via env vars

geoffjay:
Add `AGENTD_WRAP_HISTORY_BYTES` and `AGENTD_WRAP_CHANNEL_CAPACITY` environment variables to tune the PTY output ring-buffer size and broadcast channel capacity at service startup, replacing hardcoded defaults.

## Changes

- `PtyBackend::new_with_config(prefix, channel_capacity, max_history_bytes)` constructor stores and threads config into every `PtyOutputStream`
- `PtyBackend::new()` retains default behaviour (`DEFAULT_HISTORY_BYTES` / `DEFAULT_CHANNEL_CAPACITY`)
- `main.rs` reads both env vars at startup (PTY backend only) and logs the resolved values
- Doc comment in `main.rs` updated to list the new env vars
- Two new unit tests: `pty_backend_new_uses_default_buffer_config` and `pty_backend_new_with_config_applies_custom_values`

## Test plan

- [x] `cargo test -p wrap` — 135 tests pass
- [x] `cargo clippy -p wrap -- -D warnings` — clean
- [x] `cargo fmt --check -p wrap` — clean
- [x] `AGENTD_BACKEND=pty AGENTD_WRAP_HISTORY_BYTES=1048576 AGENTD_WRAP_CHANNEL_CAPACITY=512` logs resolved values on startup

Closes #1164

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1168, opened 2026-04-17
feat(orchestrator): add conversation retention policy and periodic cleanup

### LINKED ISSUE #1163: Add conversation data retention policy and cleanup
## Summary

Implement a configurable retention policy for conversation events to prevent unbounded database growth. Include automatic cleanup on agent termination and periodic pruning of old data.

## Context

With conversation events being persisted for every agent message (#1159, #1160), the orchestrator SQLite database will grow continuously. Long-running agents or agents with verbose output could generate millions of events. This issue adds lifecycle management to keep storage bounded.

## Acceptance Criteria

- [ ] New environment variables for configuration:
  - `AGENTD_CONVERSATION_RETENTION_DAYS` (default: 30) - events older than this are eligible for pruning
  - `AGENTD_CONVERSATION_MAX_EVENTS_PER_AGENT` (default: 50000) - hard cap per agent, oldest events pruned first
  - `AGENTD_CONVERSATION_CLEANUP_ON_TERMINATE` (default: false) - if true, delete all events when agent is terminated
- [ ] New storage methods on `AgentStorage`:
  - `prune_old_conversation_events(retention_days: u64) -> Result<u64>` - returns count of deleted events
  - `prune_excess_conversation_events(agent_id: Uuid, max_events: u64) -> Result<u64>`
- [ ] Periodic cleanup task (runs alongside existing reconciliation):
  - Interval: configurable, default every 6 hours
  - Applies retention_days and max_events_per_agent policies
  - Logs count of pruned events at `info` level
- [ ] On `terminate_agent()` in `crates/orchestrator/src/manager.rs`:
  - If `cleanup_on_terminate` is true, call `delete_conversation_events_for_agent()`
  - If false, events are preserved (can be viewed even for terminated agents)
- [ ] Storage metrics exposed at `/metrics`:
  - `agentd_conversation_events_total` (gauge, per agent)
  - `agentd_conversation_events_pruned_total` (counter)
  - `agentd_conversation_db_size_bytes` (gauge)

## Key Files

- `crates/orchestrator/src/storage.rs` - pruning methods
- `crates/orchestrator/src/manager.rs` - cleanup on terminate
- `crates/orchestrator/src/main.rs` - periodic cleanup task, config loading
- `crates/orchestrator/src/types.rs` - retention config types

## Blocked By
- #1159 (conversation_events table must exist)

## Stack Base
Blocked by: #1159

geoffjay:
Adds configurable retention policy for conversation events to prevent unbounded database growth. Stacked on #1167 (issue #1159).

## Changes

**`types.rs`** — `RetentionConfig` struct with `from_env()` constructor reading four env vars:
- `AGENTD_CONVERSATION_RETENTION_DAYS` (default: 30)
- `AGENTD_CONVERSATION_MAX_EVENTS_PER_AGENT` (default: 50000)
- `AGENTD_CONVERSATION_CLEANUP_ON_TERMINATE` (default: false)
- `AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS` (default: 21600 / 6 h)

**`storage.rs`** — Two new pruning methods on `AgentStorage`:
- `prune_old_conversation_events(retention_days)` — deletes rows older than the cutoff timestamp
- `prune_excess_conversation_events(agent_id, max_events)` — fetches oldest IDs beyond the cap and deletes them

**`manager.rs`** — `AgentManager` gains a `retention_config: Arc<RetentionConfig>` field. `with_retention()` constructor for explicit config (testing/main). `terminate_agent()` calls `delete_conversation_events_for_agent()` when `cleanup_on_terminate = true`.

**`main.rs`** — Reads `RetentionConfig::from_env()` at startup, passes it to `AgentManager::with_retention()`, logs resolved values, and spawns a periodic cleanup task (alongside the reconciliation loop) that runs `prune_old_conversation_events` on the configured interval and increments the `agentd_conversation_events_pruned_total` Prometheus counter.

## Test plan

- [x] 4 new unit tests: `test_prune_old_events_removes_old_rows`, `test_prune_old_events_no_op_when_nothing_old`, `test_prune_excess_removes_oldest_first`, `test_prune_excess_no_op_under_cap`
- [x] `cargo test -p orchestrator` — 417 tests pass
- [x] `cargo fmt --check -p orchestrator` — clean
- [x] `cargo clippy -p orchestrator` — no errors

Closes #1163

### DIFF
--- a/crates/orchestrator/src/main.rs
+++ b/crates/orchestrator/src/main.rs
@@ -117,9 +117,24 @@ async fn main() -> anyhow::Result<()> {
         .with_event_bus(event_bus.clone())
         .with_storage((*storage).clone());
 
+    // Conversation event retention policy (read once at startup).
+    let retention_config = Arc::new(types::RetentionConfig::from_env());
+    info!(
+        retention_days = retention_config.retention_days,
+        max_events_per_agent = retention_config.max_events_per_agent,
+        cleanup_on_terminate = retention_config.cleanup_on_terminate,
+        cleanup_interval_secs = retention_config.cleanup_interval_secs,
+        "Conversation retention config loaded",
+    );
+
     // Agent manager (Arc'd immediately so it can be shared with callbacks and API state).
-    let manager =
-        Arc::new(AgentManager::new(storage.clone(), backend, registry.clone(), ws_base_url));
+    let manager = Arc::new(AgentManager::with_retention(
+        storage.clone(),
+        backend,
+        registry.clone(),
+        ws_base_url,
+        retention_config.clone(),
+    ));
 
     // Scheduler for autonomous workflows (shares the same SeaORM connection).
     // Schema is already applied by AgentStorage::with_path() via Migrator::up().
@@ -407,6 +422,71 @@ async fn main() -> anyhow::Result<()> {
                 interval.tick().await;
                 if let Err(e) = manager_reconcile.reconcile().await {
                     error!(%e, "Periodic reconciliation failed");
+                }
+            }
+        });
+    }
+
+    // Periodic conversation event cleanup: prune old events and enforce the
+    // per-agent cap on a configurable interval (default 6 h).
+    {
+        let storage_cleanup = storage.clone();
+        let retention = retention_config.clone();
+        let cleanup_interval = retention.cleanup_interval_secs;
+        info!(
+            cleanup_interval_secs = cleanup_interval,
+            "Starting periodic conversation cleanup loop"
+        );
+        tokio::spawn(async move {
+            let mut interval =
+                tokio::time::interval(tokio::time::Duration::from_secs(cleanup_interval));
+            interval.tick().await; // skip immediate tick on startup
+            loop {
+                interval.tick().await;
+
+                // Age-based pruning: remove events older than retention_days.
+                match storage_cleanup.prune_old_conversation_events(retention.retention_days).await
+                {
+                    Ok(n) if n > 0 => {
+                        info!(pruned = n, "Pruned old conversation events");
+                        metrics::counter!("agentd_conversation_events_pruned_total").increment(n);
+                    }
+                    Ok(_) => {}
+                    Err(e) => error!(%e, "Conversation event age-prune failed"),
+                }
+
+                // Per-agent cap: enforce max_events_per_agent for every known agent.
+                match storage_cleanup.list(None, None).await {
+                    Ok(agents) => {
+                        for agent in &agents {
+                            match storage_cleanup
+                                .prune_excess_conversation_events(
+                                    agent.id,
+                                    retention.max_events_per_agent,
+                                )
+                                .await
+                            {
+                                Ok(n) if n > 0 => {
+                                    info!(
+                                        agent_id = %agent.id,
+                                        pruned = n,
+                                        "Pruned excess conversation events for agent",
+                                    );
+                                    metrics::counter!("agentd_conversation_events_pruned_total")
+                                        .increment(n);
+                                }
+                                Ok(_) => {}
+                                Err(e) => {
+                                    error!(
+                                        agent_id = %agent.id,
+                                        %e,
+                                        "Conversation event cap-prune failed for agent",
+                                    );
+                                }
+                            }
+                        }
+                    }
+                    Err(e) => error!(%e, "Failed to list agents for per-agent cap pruning"),
                 }
             }
         });
--- a/crates/orchestrator/src/manager.rs
+++ b/crates/orchestrator/src/manager.rs
@@ -1,6 +1,8 @@
 use crate::scheduler::events::SystemEvent;
 use crate::storage::{AgentStorage, ProjectStorage};
-use crate::types::{Agent, AgentConfig, AgentStatus, AgentUsageStats, ClearContextResponse};
+use crate::types::{
+    Agent, AgentConfig, AgentStatus, AgentUsageStats, ClearContextResponse, RetentionConfig,
+};
 use crate::websocket::ConnectionRegistry;
 use chrono::Utc;
 use std::sync::Arc;
@@ -21,6 +23,8 @@ pub struct AgentManager {
     registry: ConnectionRegistry,
     /// The base URL agents will use to connect back via WebSocket.
     ws_base_url: String,
+    /// Conversation event retention policy applied at terminate and on schedule.
+    retention_config: Arc<RetentionConfig>,
 }
 
 impl AgentManager {
@@ -30,7 +34,25 @@ impl AgentManager {
         registry: ConnectionRegistry,
         ws_base_url: String,
     ) -> Self {
-        Self { storage, backend, registry, ws_base_url }
+        Self::with_retention(
+            storage,
+            backend,
+            registry,
+            ws_base_url,
+            Arc::new(RetentionConfig::from_env()),
+        )
+    }
+
+    /// Construct with an explicit [`RetentionConfig`] (useful for testing and
+    /// for wiring up the config read from env vars in `main`).
+    pub fn with_retention(
+        storage: Arc<AgentStorage>,
+        backend: Arc<dyn ExecutionBackend>,
+        registry: ConnectionRegistry,
+        ws_base_url: String,
+        retention_config: Arc<RetentionConfig>,
+    ) -> Self {
+        Self { storage, backend, registry, ws_base_url, retention_config }
     }
 
     pub fn registry(&self) -> &ConnectionRegistry {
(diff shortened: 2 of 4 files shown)


geoffjay (commented):
## Review: feat(orchestrator): add conversation retention policy and periodic cleanup

**Stack note:** This PR is stacked on `issue-1159` (PR #1167), which currently has `needs-rework`. These changes cannot merge until #1167 is fixed and this branch is restacked on the updated parent. The code review below is independent of that dependency.

---

### 🔴 Fix required — wrong Prometheus counter method (`.absolute` vs `.increment`)

**`main.rs`, periodic cleanup task:**

```rust
// Current — WRONG
metrics::counter!("agentd_conversation_events_pruned_total").absolute(n);
```

`.absolute(n)` *sets* the counter to `n`, overwriting the accumulated total on every prune run. After two prune cycles of 100 events each the counter reads `100`, not `200`.

Every other counter in the codebase uses `.increment()`:

```rust
// api.rs
metrics::counter!("agents_created_total").increment(1);
metrics::counter!("agents_terminated_total").increment(1);

// message_bridge.rs — .absolute(0) only used for initialisation to zero at startup
metrics::counter!("messages_delivered_to_agents").increment(1);

// main.rs (existing)
metrics::counter!("context_clears_total", "trigger" => "auto").increment(1);
```

Fix:

```rust
metrics::counter!("agentd_conversation_events_pruned_total").increment(n);
```

---

### 🔴 Fix required — zero `cleanup_interval_secs` panics at startup

**`types.rs`, `RetentionConfig::from_env()`:**

```rust
cleanup_interval_secs: std::env::var("AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS")
    .ok()
    .and_then(|v| v.parse().ok())
    .unwrap_or(21_600),
```

If the variable is set to `"0"`, `parse::<u64>().ok()` returns `Some(0)`, so `unwrap_or(21_600)` never fires. Then:

```rust
tokio::time::interval(tokio::time::Duration::from_secs(0))  // panics: "Period is zero"
```

This is the same class of bug flagged in PR #1166 (`AGENTD_WRAP_CHANNEL_CAPACITY=0`). Fix with a validation guard:

```rust
cleanup_interval_secs: std::env::var("AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS")
    .ok()
    .and_then(|v| v.parse().ok())
    .filter(|&v| v >= 60)          // reject values too small to be intentional
    .unwrap_or(21_600),
```

Or fail fast at startup:

```rust
let interval_secs: u64 = std::env::var("AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS")
    .ok()
    .and_then(|v| v.parse().ok())
    .unwrap_or(21_600);
anyhow::ensure!(interval_secs > 0, 
    "AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS must be > 0, got {}", interval_secs);
```

Same validation should be applied to `retention_days` (zero would delete ALL events — cutoff becomes `now - 0 days = now`) and `max_events_per_agent` (zero would delete all events for every connected agent).

---

### 🟡 Important gap — `prune_excess_conversation_events` is never called

`max_events_per_agent` is configured, logged at startup, and the pruning method exists — but the periodic cleanup task only calls `prune_old_conversation_events`. The per-agent cap is never enforced:

```rust
// main.rs cleanup loop — only age-based pruning fires
match storage_cleanup.prune_old_conversation_events(retention.retention_days).await {
    // prune_excess_conversation_events is never called here
```

If this is intentional (cap enforcement is on-terminate only), the doc comment on `RetentionConfig::max_events_per_agent` should say so:

```rust
/// Hard cap per agent; oldest events are evicted first **on agent termination when
/// `cleanup_on_terminate = true`**. The periodic cleanup task only enforces `retention_days`.
/// (env: `AGENTD_CONVERSATION_MAX_EVENTS_PER_AGENT`, default: 50000).
pub max_events_per_agent: u64,
```

If the cap *should* be enforced periodically, the loop needs to fetch all agent IDs from storage and call `prune_excess_conversation_events` for each, similar to how the reconciliation loop works.

---

### ✅ What's correct

- **FK ordering in `terminate_agent()`**: `delete_conversation_events_for_agent()` at line 267 fires *before* `storage.delete(id)` at line 278 — correct order, no orphaned events
- **Builder pattern**: `with_retention()` follows the existing `with_event_bus()` pattern; `new()` correctly delegates to it with env-derived defaults
- **Cleanup task structure**: `interval.tick().await` before the loop (skip startup), then `loop { interval.tick().await; ... }` — matches the reconciliation loop pattern exactly
- **Error handling**: cleanup errors logged at `warn`, don't interrupt the prune loop or the termination path
- **Test coverage**: 4 tests correctly exercise old-row pruning, no-op on nothing old, excess pruning (oldest-first), and no-op under cap — all with cross-agent isolation
- **`prune_excess_conversation_events` two-phase approach**: fetch IDs then delete by `is_in()` is the correct SQLite workaround for correlated subqueries

ShawnSunClio (changes requested):
## Review — feat(orchestrator): add conversation retention policy and periodic cleanup

**Stack note:** This PR is stacked on #1167 (`issue-1159 → issue-1163`), which is still open. The diff is clean relative to its parent — no merge conflicts detected — but this branch cannot be merged until #1167 lands first.

---

### Blocking issues (3)

#### 1. `metrics::counter!(...).absolute(n)` resets the counter instead of incrementing it (`main.rs:447`)

```rust
// current — WRONG
metrics::counter!("agentd_conversation_events_pruned_total").absolute(n);
// fix
metrics::counter!("agentd_conversation_events_pruned_total").increment(n);
```

`Counter::absolute(n)` sets the counter to `n`; it does not add `n` to the running total. Every cleanup cycle would overwrite the counter with the number pruned in that single run. After two cycles pruning 100 and 50 events respectively, the counter would read 50, not 150. Every other counter in this file uses `.increment()`; this should too.

---

#### 2. `max_events_per_agent` is configured and logged but never enforced

`AGENTD_CONVERSATION_MAX_EVENTS_PER_AGENT` is documented in the PR description as a working config knob, and `RetentionConfig` logs its value at startup. But the periodic cleanup loop (main.rs ~line 440–450) only calls `prune_old_conversation_events` — `prune_excess_conversation_events` is never invoked at the service level. The per-agent cap is dead code at runtime.

Either wire `prune_excess_conversation_events` into the cleanup loop (iterating over active agents), or explicitly document that this field is reserved for a future phase and remove it from the startup log until it is wired up. Logging a value that has no runtime effect is misleading.

---

#### 3. `cleanup_interval_secs = 0` panics the service (`main.rs:439`)

```rust
tokio::time::interval(tokio::time::Duration::from_secs(cleanup_interval))
```

`tokio::time::interval` panics with *"period must be positive"* when given `Duration::ZERO`. A user who sets `AGENTD_CONVERSATION_CLEANUP_INTERVAL_SECS=0` (or an env var that parses to 0) will crash the orchestrator at startup. Add a minimum bound — either clamp to `max(cleanup_interval, 1)` or return an explicit startup error with a descriptive message.

---

### Non-blocking suggestions

#### 4. `impl Default` delegates to `from_env()` — surprising in tests (`types.rs:2137`)

```rust
impl Default for RetentionConfig {
    fn default() -> Self { Self::from_env() }
}
```

`Default` conventionally yields hardcoded safe values with no side effects. Delegating to `from_env()` means any test that constructs `AgentManager::new()` (which calls `RetentionConfig::from_env()` internally) will silently inherit whatever env vars are set in the test runner. The better pattern is to implement `Default` with hardcoded values and have `from_env()` build from `Self::default()` as a baseline — that way tests get predictable defaults unless they explicitly set env vars.

#### 5. `test_prune_excess_removes_oldest_first` may be flaky on fast machines (`storage.rs:1857`)

Five events are inserted in a tight loop using `ConversationEvent::new()`, which stamps each with `Utc::now()`. On a fast machine, multiple inserts may land within the same millisecond, giving identical `created_at` values. `ORDER BY created_at ASC` is then non-deterministic, and the assertion `sessions.iter().all(|&s| s >= 2)` could fail intermittently.

Fix by constructing events with explicitly distinct timestamps:

```rust
for i in 0..5i64 {
    let mut ev = ConversationEvent::new(...);
    ev.created_at = Utc::now() - chrono::Duration::seconds(5 - i); // oldest first
    storage.insert_conversation_event(&ev).await.unwrap();
}
```

#### 6. `retention_days = 0` silently deletes everything (`storage.rs:prune_old_conversation_events`)

`prune_old_conversation_events(0)` computes `cutoff = Utc::now() - Duration::days(0)` ≈ `Utc::now()`, which deletes all events created before the current instant — effectively wiping the table. A failed parse falling back to the default of 30 is fine, but a user who intentionally or accidentally sets the variable to `0` gets silent data loss. Worth a guard (`if retention_days == 0 { return Ok(0); }`) or a startup validation error.

---

### Summary

Three issues require fixes before merge: the counter semantic (`absolute` → `increment`), the unconnected per-agent cap, and the zero-interval panic. Suggestions 4–6 are lower priority but worth addressing to avoid future pain.

geoffjay:

- #1167
    - #1168 ◀

PR #1170, opened 2026-04-17
feat(orchestrator): add REST API endpoints for conversation history

### LINKED ISSUE #1161: Add REST API endpoints for conversation history retrieval
## Summary

Add REST endpoints to the orchestrator API for retrieving persisted conversation history for an agent. These endpoints enable the UI (and CLI) to load full conversation history on demand, independent of live WebSocket streaming.

## Context

Once conversation events are persisted (#1159, #1160), consumers need a way to retrieve them. The orchestrator already serves agent data at `GET /agents/{id}` and usage at `GET /agents/{id}/usage`. This adds a conversation history endpoint following the same patterns.

## Acceptance Criteria

- [ ] New endpoint: `GET /agents/{id}/conversation` returning paginated conversation events
- [ ] Query parameters:
  - `limit` (default: 200, max: 1000) - number of events to return
  - `before` (RFC3339 timestamp) - cursor for backward pagination
  - `after` (RFC3339 timestamp) - cursor for forward pagination (used for "load newer")
  - `event_type` (optional, comma-separated) - filter by event types (e.g., `output,tool_use,thinking`)
  - `session` (optional, integer) - filter by session number
- [ ] Response format matches the stream event format for easy UI reuse:
  ```json
  {
    "events": [
      {
        "id": "uuid",
        "type": "agent:output",
        "agent_id": "uuid",
        "line": "text content",
        "timestamp": "2026-04-17T...",
        "session_number": 0
      }
    ],
    "total": 1234,
    "has_more": true
  }
  ```
- [ ] New endpoint: `GET /agents/{id}/conversation/summary` returning event counts by type and session boundaries
- [ ] New endpoint: `DELETE /agents/{id}/conversation` to clear conversation history for an agent
- [ ] Route registered in `create_router()` in `crates/orchestrator/src/api.rs`
- [ ] Corresponding client methods added to `crates/orchestrator/src/client.rs`:
  - `get_conversation_history(id, query) -> Result<ConversationHistoryResponse>`
  - `get_conversation_summary(id) -> Result<ConversationSummary>`
  - `delete_conversation_history(id) -> Result<()>`
- [ ] Events returned in ascending `created_at` order (oldest first) for chronological display

## Key Files

- `crates/orchestrator/src/api.rs` - add route and handler functions
- `crates/orchestrator/src/client.rs` - add client methods
- `crates/orchestrator/src/storage.rs` - uses query methods from #1159
- `crates/orchestrator/src/types.rs` - add response types

## Blocked By
- #1159 (conversation_events table and storage layer must exist first)

## Stack Base
Blocked by: #1159

geoffjay:
Implements three REST API endpoints for retrieving persisted conversation events, backed by the storage layer added in #1159.

## Endpoints added

- `GET /agents/{id}/conversation` - paginated event list with `limit`, `before`/`after`, `event_type`, and `session` filters
- `GET /agents/{id}/conversation/summary` - aggregate statistics (total events, per-type counts, session count, first/last timestamps)
- `GET /agents/{id}/conversation/{event_id}` - single event by ID (404 if not found or wrong agent)

## Changes

- `types.rs`: `ConversationEventResponse`, `ConversationHistoryQuery`, `ConversationHistoryResponse`, `ConversationSummary` response types
- `storage.rs`: `get_conversation_event_by_id` and `get_conversation_summary` methods + 4 new unit tests
- `api.rs`: 3 handler functions + routes registered in the router
- `client.rs`: `list_conversation_events`, `get_conversation_summary`, `get_conversation_event` typed client helpers

Closes #1161

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4944,6 +4944,7 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-subscriber",
+ "urlencoding",
  "uuid",
  "wiremock",
  "wrap",
--- a/crates/orchestrator/Cargo.toml
+++ b/crates/orchestrator/Cargo.toml
@@ -75,4 +75,5 @@ libc = "0.2"
 [dev-dependencies]
 tempfile = "3.14"
 tower = { version = "0.5", features = ["util"] }
+urlencoding = "2"
 wiremock = "0.6"
--- a/crates/orchestrator/Cargo.toml
+++ b/crates/orchestrator/Cargo.toml
@@ -43,6 +43,9 @@ metrics-exporter-prometheus = { workspace = true }
 # default-features = false avoids native-tls (macOS TLS init panics on non-main threads).
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
 
+# URL encoding for query parameters
+urlencoding = "2.1"
+
 # WebSocket client (for MessageBridge → communicate service)
 tokio-tungstenite = { version = "0.26", features = ["rustls-tls-webpki-roots"] }
 
--- a/crates/orchestrator/src/api.rs
+++ b/crates/orchestrator/src/api.rs
@@ -74,6 +74,10 @@ pub fn create_router(state: ApiState) -> Router {
         .route("/agents/{id}/usage", get(get_agent_usage))
         .route("/agents/{id}/clear-context", post(clear_agent_context))
         .route("/agents/{id}/approvals", get(list_agent_approvals))
+        // Conversation history (static sub-path must precede wildcard)
+        .route("/agents/{id}/conversation", get(get_agent_conversation))
+        .route("/agents/{id}/conversation/summary", get(get_agent_conversation_summary))
+        .route("/agents/{id}/conversation/{event_id}", get(get_agent_conversation_event))
         .route("/agents/{id}/rooms", get(list_agent_rooms).post(join_agent_room))
         .route("/agents/{id}/rooms/{room_id}", axum::routing::delete(leave_agent_room))
         .route(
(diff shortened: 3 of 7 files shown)


geoffjay (commented):
## Review: feat(orchestrator): add REST API endpoints for conversation history

**Stack note:** This PR is stacked on `issue-1159` (PR #1167, `needs-rework`). It cannot merge until #1167 is fixed and restacked. The code review below is independent of that dependency.

---

### 🔴 Fix required — in-memory session filter breaks cursor pagination

**`api.rs`, `get_agent_conversation`:**

```rust
// Fetch limit+1 events with NO session filter at the DB level
let opts = ConversationQuery {
    event_types,
    since,
    until,
    limit: Some(limit + 1),  // no session_number field
    offset: None,
};
let mut events = state.manager.agent_storage()
    .list_conversation_events(id, &opts).await?;

// Session filter applied in memory AFTER the DB-limited fetch
if let Some(session) = query.session {
    events.retain(|e| e.session_number == session);
}

let has_more = events.len() as u64 > limit;
```

**This is incorrect for the pagination case.** `list_conversation_events` returns events ordered by `created_at ASC`. Sessions are sequential — session 0 events are always older than session 1 events. Consider a request with `limit=5&session=1` against an agent that has 90 events in session 0 and 10 in session 1:

1. DB fetches rows 1–6 (all session 0, because they're oldest)
2. `retain` removes all 6 (none match session 1)
3. Result: empty list, `has_more = false`
4. **But 10 session-1 events exist and were never seen**

The `before`/`after` cursor fields compound the problem: a client cannot page through session-filtered results at all because the DB-level fetch doesn't know about the session boundary.

**Fix:** Push the session filter to the DB query. This requires adding `session_number: Option<i64>` to `ConversationQuery` (recommended in the #1159 pre-implementation analysis, but not implemented in #1167). Until #1167 exposes that field, the `session` query parameter should either be removed from this endpoint or documented as "approximate / best-effort, do not use with pagination."

Minimal fix for this PR, assuming `ConversationQuery` gains the field via the rework of #1167:

```rust
let opts = ConversationQuery {
    event_types,
    since,
    until,
    session_number: query.session,  // pushed to DB
    limit: Some(limit + 1),
    offset: None,
};
// remove the in-memory retain block entirely
```

---

### 🔴 Fix required — `get_conversation_summary` loads all events into memory

**`storage.rs`, `get_conversation_summary`:**

```rust
pub async fn get_conversation_summary(&self, agent_id: Uuid) -> Result<ConversationSummary> {
    // No limit — fetches every event for this agent
    let events = self.list_conversation_events(agent_id, &ConversationQuery::default()).await?;
    // … iterates to compute counts
}
```

With the default retention cap of 50,000 events per agent, this loads up to 50 k `ConversationEvent` structs into a `Vec`, then iterates over them. That's O(N) memory and O(N) CPU where N ≤ 50,000 on every call to `GET /agents/{id}/conversation/summary`.

The correct approach is SQL aggregates. SeaORM supports raw statements:

```rust
use sea_orm::{ConnectionTrait, Statement, DbBackend};

// Per-type counts + first/last timestamps in one query
let rows = self.db.query_all(Statement::from_sql_and_values(
    DbBackend::Sqlite,
    r#"SELECT event_type, COUNT(*) as cnt,
              MIN(created_at) as first_at, MAX(created_at) as last_at
       FROM conversation_events WHERE agent_id = ? GROUP BY event_type"#,
    [agent_id.to_string().into()],
)).await?;

// Distinct session count in a second query
let session_count: u64 = self.db.query_one(Statement::from_sql_and_values(
    DbBackend::Sqlite,
    "SELECT COUNT(DISTINCT session_number) FROM conversation_events WHERE agent_id = ?",
    [agent_id.to_string().into()],
)).await?
    .and_then(|r| r.try_get_by_index::<i64>(0).ok())
    .unwrap_or(0) as u64;
```

Two DB round-trips instead of one full table scan loaded into Rust memory. The current implementation would be unacceptably slow for agents with large histories.

---

### 🟡 Inconsistent limit cap — should use `clamp_limit()` or document the override

**`api.rs`, `get_agent_conversation`:**

```rust
let limit = query.limit.unwrap_or(100).min(500);
```

Every other paginated endpoint in `api.rs` uses the shared `clamp_limit()` helper (which caps at `MAX_PAGE_LIMIT = 200`):

```rust
// list_agent_approvals, list_projects, list_project_agents, list_project_workflows…
let limit = clamp_limit(query.limit);
```

This endpoint bypasses that cap with its own `.min(500)`. If a conversation limit of 500 is intentional (conversation replay needs larger pages than typical list endpoints), add a constant and document the exception. If it should follow the standard cap, use `clamp_limit`.

---

### 🟡 `total` in response is unfiltered agent count

**`api.rs`, `get_agent_conversation`:**

```rust
let total = state.manager.agent_storage()
    .count_conversation_events(id)  // counts ALL events, ignores filters
    .await
    .map_err(ApiError::Internal)?;

Ok(Json(ConversationHistoryResponse { events: items, total, has_more }))
```

When `event_type=output` or `session=1` is specified, `total` still returns the full count for all event types and all sessions. A client trying to render "showing 5 of N matching events" will show an incorrect denominator.

If `total` is intended to mean "total events for this agent regardless of filter," the field should be renamed `total_events` and documented as unfiltered. If it should reflect the filtered count, it needs a separate `count_conversation_events` variant that accepts filter options.

---

### ✅ What's correct

- Route ordering: `/conversation/summary` declared before `/conversation/{event_id}` — prevents ambiguity ✅
- Agent existence check on all three handlers before storage calls ✅
- `get_conversation_event_by_id` enforces agent ownership at the DB level (filters on both `agent_id` and `id`) ✅
- `has_more` via `limit + 1` trick is correct (when the session filter is removed from the in-memory path) ✅
- `ConversationEventResponse::from` correctly prefixes `event_type` with `"agent:"` to match WebSocket stream format ✅
- Client methods correctly construct query strings; `get_conversation_summary` and `get_conversation_event` return typed responses ✅
- 4 storage tests cover: correct fetch, wrong-agent isolation (returns None), empty summary, count+session aggregation ✅

ShawnSunClio (changes requested):
## Review — feat(orchestrator): add REST API endpoints for conversation history

**Stack note:** This PR bases on `issue-1159` (PR #1167, still open). It is a **sibling** of PR #1168 — both depend on #1167 landing first. The diff is conflict-free relative to the parent branch.

---

### Blocking issues (3)

#### 1. `get_conversation_summary` loads the entire event table into memory (`storage.rs`)

```rust
// Current — fetches ALL events for the agent with no limit
let events = self.list_conversation_events(agent_id, &ConversationQuery::default()).await?;
```

`ConversationQuery::default()` has `limit: None`, which means `list_conversation_events` calls `.all(&self.db)` with no LIMIT clause. For an agent at the 50 k default cap, every `/conversation/summary` request loads 50 k rows into the Rust process just to count types and compute min/max timestamps. This will be the hottest endpoint once the table grows.

Replace with SQL aggregation:

```rust
// Approach: run two targeted queries instead of a full scan
// 1. SELECT event_type, COUNT(*) FROM conversation_events WHERE agent_id = ?
//    GROUP BY event_type
// 2. SELECT COUNT(DISTINCT session_number), MIN(created_at), MAX(created_at),
//    COUNT(*) FROM conversation_events WHERE agent_id = ?
```

Both are expressible with SeaORM's `select_only()` / `column_as()` / `group_by()` / `PaginatorTrait::count()` APIs and avoid materializing any event rows.

---

#### 2. Session filter breaks `has_more` and `total` — pagination is semantically wrong (`api.rs`)

The handler fetches `limit + 1` events from the database with no session filter, then applies the session filter in memory:

```rust
// DB fetches up to 101 unfiltered events...
let opts = ConversationQuery { limit: Some(limit + 1), .. };
let mut events = storage.list_conversation_events(id, &opts).await?;

// ...then session filter is applied in memory:
if let Some(session) = query.session {
    events.retain(|e| e.session_number == session);
}

// has_more and total are now wrong:
let has_more = events.len() as u64 > limit;  // based on post-filter count — incorrect
let total = storage.count_conversation_events(id).await?;  // counts ALL events, not session
```

**`has_more` is wrong:** Suppose the agent has 200 events: 5 in session 3, 195 in sessions 0–2 (which are chronologically earlier). A request for `limit=100&session=3` fetches the 101 oldest events, which are all from sessions 0–2. After session filtering: 0 results, `has_more = false`. But session 3's 5 events exist — they're just not in the first 101 rows. The caller gets an empty page with no way to paginate to the actual session 3 events.

Even in a simpler case where session events are evenly distributed, `has_more` is computed on the post-filter slice of a pre-filter window, which is incorrect for cursor pagination.

**`total` is wrong:** `count_conversation_events` returns the total count across all sessions. A caller using `total` to display "showing X of Y" would show the wrong denominator.

**Fix:** Add `session_number: Option<i64>` to `ConversationQuery` and push the filter to SQL — the same way `event_types`, `since`, and `until` are handled. Then `has_more` and `total` will be computed on the correctly-filtered result set.

---

#### 3. Client query string does not URL-encode timestamp values — `+` is corrupted (`client.rs`)

```rust
if let Some(ref before) = query.before {
    params.push(format!("before={before}"));  // NOT encoded
}
```

RFC 3339 timestamps with non-UTC timezone offsets contain `+`, e.g. `2026-04-17T09:00:00+05:00`. In a URL query string `+` is decoded as a space (application/x-www-form-urlencoded encoding), so the server receives `before=2026-04-17T09:00:00 05:00`, which fails `DateTime::parse_from_rfc3339` and returns `ApiError::InvalidInput`.

Fix: percent-encode the values. The simplest approach is to add `urlencoding` (already a common transitive dep) and call `urlencoding::encode(before)`, or restrict `before`/`after` to UTC-only (`Z` suffix) in the client docs and add a validation check. Either way the current client is broken for non-UTC callers.

---

### Non-blocking observations

#### 4. Route ordering comment is correct — no change needed

The comment `// static sub-path must precede wildcard` correctly explains why `/conversation/summary` is registered before `/conversation/{event_id}`. Axum resolves static segments before wildcards at the same path depth, so the ordering is right and the comment is a useful maintenance note.

#### 5. `get_conversation_event_by_id` agent-scoping is correct

The double-condition filter (`AgentId = ? AND Id = ?`) correctly prevents agent A from reading agent B's events by ID, returning `None` → 404. Tests cover both the happy path and the cross-agent isolation case.

#### 6. `ConversationHistoryQuery.session` type matches `ConversationEvent.session_number` (`i64`)

Consistent — no mismatch between query param type and storage type.

---

### Summary

Three fixes needed before merge:
1. Replace the full-scan `get_conversation_summary` with SQL aggregation queries
2. Push the `session_number` filter into `ConversationQuery` so `has_more`/`total` are accurate
3. URL-encode `before`/`after` values in the client query-string builder

geoffjay:

- #1167
    - #1170 ◀
        - #1175

Issue #1162, opened 2026-04-17
Load persisted conversation history in UI agent detail view


geoffjay:
## Summary

Update the UI agent detail page to load persisted conversation history from the orchestrator REST API on mount, then seamlessly transition to the live WebSocket stream. This is the core user-facing change - entering an agent view will show full history even if output was generated in the background.

## Context

Currently, `useAgentStream.ts` connects to `ws://.../stream/{agentId}` and only receives events from that point forward. Historical output from before the connection is lost (except for up to 1,000 lines in `sessionStorage`). With the new `GET /agents/{id}/conversation` endpoint (#1161), the UI can load the complete history first.

## Acceptance Criteria

- [ ] New service method in `ui/src/services/orchestrator.ts`:
  - `getConversationHistory(agentId, params)` calling `GET /agents/{id}/conversation`
  - `getConversationSummary(agentId)` calling `GET /agents/{id}/conversation/summary`
- [ ] New hook `ui/src/hooks/useConversationHistory.ts` (or integrated into `useAgentStream`):
  - On mount: fetch persisted history via REST API (latest N events)
  - Convert API response events to `LogLine[]` format (reuse existing `makeLogLine`, `makeToolUseLine`, `makeThinkingLine`)
  - After history loads, connect to WebSocket stream for live updates
  - Deduplicate events that arrive via both REST and WebSocket (by timestamp + content hash or event ID)
- [ ] Loading state shown while history is being fetched ("Loading conversation history...")
- [ ] Smooth transition: no visual jump when switching from historical to live data
- [ ] Infinite scroll upward to load older history (using `before` cursor pagination)
- [ ] Session boundaries rendered as visual separators (when `context_cleared` events are encountered)
- [ ] `sessionStorage` persistence retained as fallback cache but REST API is the primary source
- [ ] Empty state: if no conversation history exists, show appropriate message then live stream

## Implementation Strategy

```
Mount AgentDetail
    |
    v
Fetch GET /agents/{id}/conversation?limit=200 (latest 200 events)
    |
    v
Render historical LogLines
    |
    v
Connect WebSocket /stream/{agentId}
    |
    v
Deduplicate: skip WS events with timestamp <= last historical event
    |
    v
Append live events normally
```

### Deduplication Approach

The REST API returns events with UUIDs. The WebSocket stream events do not currently include event IDs. Options:
1. **Timestamp-based**: skip WS events with `timestamp <= lastHistoricalTimestamp` (simple, may miss edge cases)
[...]

PR #1172, opened 2026-04-17
feat(orchestrator): persist agent stream events in WebSocket handler

### LINKED ISSUE #1160: Persist agent stream events in orchestrator WebSocket handler
## Summary

Wire the new `conversation_events` storage layer into the orchestrator's WebSocket message handler so that every agent event (output, tool use, thinking, result) is persisted to SQLite as it flows through.

## Context

The `handle_incoming_message()` function in `crates/orchestrator/src/websocket.rs` already parses every NDJSON message from Claude Code agents and broadcasts structured events to `/stream` subscribers. This issue adds a parallel async write to the conversation_events table for each event, without blocking the broadcast path.

## Acceptance Criteria

- [ ] `handle_incoming_message()` receives an `AgentStorage` reference (via `ConnectionRegistry` or function parameter)
- [ ] Each broadcast event also triggers an async `insert_conversation_event()` call
- [ ] Event types persisted:
  - `assistant` messages -> `output` events (one per text line), `tool_use` events, `thinking` events
  - `result` messages -> `result` events (with usage metadata)
  - `control_request` -> not persisted (approval flow is separate)
  - `system` -> not persisted (lifecycle only)
- [ ] `prompt_sent` events persisted when `send_message_to_agent()` is called (from `ConnectionRegistry::send_message`)
- [ ] `activity_changed` events persisted on Idle/Busy transitions
- [ ] `context_cleared` events persisted with session_number increment
- [ ] Storage writes are fire-and-forget (spawned task) to avoid blocking the WebSocket read loop
- [ ] Errors in storage writes are logged at `warn` level but do not interrupt message flow
- [ ] Session number tracked per-agent and incremented on `context_cleared` events

## Implementation Notes

The `ConnectionRegistry` already holds in-memory state. Add an `Option<AgentStorage>` field:

```rust
pub struct ConnectionRegistry {
    // ... existing fields ...
    storage: Option<AgentStorage>,  // NEW
    session_numbers: Arc<RwLock<HashMap<Uuid, i64>>>,  // NEW - per-agent session counter
}
```

In `handle_incoming_message()`, after each broadcast, spawn a storage write:

```rust
if let Some(ref storage) = registry.storage {
    let storage = storage.clone();
    let event = ConversationEvent { /* ... */ };
    tokio::spawn(async move {
        if let Err(e) = storage.insert_conversation_event(&event).await {
            warn!(%agent_id, %e, "Failed to persist conversation event");
        }
    });
}
```

## Key Files

- `crates/orchestrator/src/websocket.rs` - main integration point
- `crates/orchestrator/src/storage.rs` - storage methods (from #1159)
- `crates/orchestrator/src/types.rs` - ConversationEvent types (from #1159)
- `crates/orchestrator/src/main.rs` - pass storage to ConnectionRegistry

## Blocked By
- #1159 (conversation_events table and storage layer must exist first)

## Stack Base
Blocked by: #1159

geoffjay:
Wires `AgentStorage` into `ConnectionRegistry` so that every agent event flowing through the WebSocket handler is persisted to SQLite as a `conversation_event` row, without blocking the WebSocket read loop.

### `websocket.rs`
- Added `storage: Option<AgentStorage>` and `session_numbers: Arc<RwLock<HashMap<Uuid, i64>>>` to `ConnectionRegistry`
- `with_storage(storage)` builder method to attach storage at startup
- `persist_event()` — fire-and-forget helper; spawns a task, logs warn on error, never panics
- `get_session_number()` — reads the in-memory per-agent session counter
- `persist_context_cleared()` — updates the session counter, persists the event, and broadcasts `agent:context_cleared` on the stream
- `register()` — initialises session counter to 0 then fires an async lookup to sync from storage
- `unregister()` — cleans up the session counter entry
- `send_user_message()` — persists `prompt_sent` + `activity_changed(busy)`
- `handle_incoming_message()` — persists `output` (per line), `tool_use`, `thinking`, `result` with usage metadata, `activity_changed(idle)`; also fixes the hardcoded `session_number: 0` in the `usage_update` stream event

### `manager.rs`
- `clear_context()` calls `registry.persist_context_cleared()` after advancing the storage session

### `main.rs`
- Registry construction chains `.with_storage((*storage).clone())`

## Event types persisted
| Message type | Events written |
|---|---|
| `assistant` | `output` (per line), `tool_use`, `thinking` |
| `result` | `result` (with usage metadata), `activity_changed(idle)` |
| `send_user_message` | `prompt_sent`, `activity_changed(busy)` |
| `clear_context` | `context_cleared` |
| `system`, `control_request`, `keep_alive` | not persisted |

Closes #1160

### DIFF
--- a/crates/orchestrator/src/main.rs
+++ b/crates/orchestrator/src/main.rs
@@ -111,8 +111,11 @@ async fn main() -> anyhow::Result<()> {
     // Shared event bus for internal lifecycle events.
     let event_bus = EventBus::shared(256);
 
-    // WebSocket connection registry.
-    let registry = ConnectionRegistry::new().with_event_bus(event_bus.clone());
+    // WebSocket connection registry — attach storage so conversation events are
+    // persisted as they flow through the WebSocket handler.
+    let registry = ConnectionRegistry::new()
+        .with_event_bus(event_bus.clone())
+        .with_storage((*storage).clone());
 
     // Agent manager (Arc'd immediately so it can be shared with callbacks and API state).
     let manager =
--- a/crates/orchestrator/src/manager.rs
+++ b/crates/orchestrator/src/manager.rs
@@ -660,6 +660,10 @@ impl AgentManager {
             bus.publish(SystemEvent::ContextCleared { agent_id: *id });
         }
 
+        // Persist context_cleared conversation event and advance the in-memory
+        // session counter so subsequent events use the new session number.
+        self.registry.persist_context_cleared(*id, new_session_number as i64).await;
+
         info!(agent_id = %id, new_session_number, "Agent context cleared");
 
         Ok(ClearContextResponse { agent_id: *id, session_usage, new_session_number })
--- a/crates/orchestrator/src/storage.rs
+++ b/crates/orchestrator/src/storage.rs
@@ -17,8 +17,8 @@ use crate::{
 use anyhow::Result;
 use chrono::{DateTime, Utc};
 use sea_orm::{
-    ColumnTrait, Condition, DatabaseConnection, EntityTrait, Order, PaginatorTrait, QueryFilter,
-    QueryOrder, QuerySelect, Set,
+    ColumnTrait, Condition, ConnectionTrait, DatabaseConnection, EntityTrait, Order,
+    PaginatorTrait, QueryFilter, QueryOrder, QuerySelect, Set, Statement,
 };
 use sea_orm_migration::prelude::MigratorTrait;
 use std::collections::HashMap;
@@ -667,6 +667,26 @@ impl AgentStorage {
             .count(&self.db)
             .await?;
         Ok(count)
+    }
+
+    /// Returns the highest `session_number` stored in `conversation_events` for
+    /// `agent_id`, or `0` if no events exist yet.
+    ///
+    /// Used by [`crate::websocket::ConnectionRegistry::register`] to restore the
+    /// correct session counter on agent reconnect - reading directly from the
+    /// events table avoids any dependency on usage-session semantics.
+    pub async fn get_max_conversation_session_number(&self, agent_id: Uuid) -> Result<i64> {
+        let row = self
+            .db
+            .query_one(Statement::from_sql_and_values(
+                self.db.get_database_backend(),
+                r#"SELECT COALESCE(MAX(session_number), 0) AS max_session
+               FROM conversation_events
+               WHERE agent_id = ?"#,
+                [sea_orm::Value::String(Some(Box::new(agent_id.to_string())))],
+            ))
+            .await?;
+        Ok(row.and_then(|r| r.try_get::<i64>("", "max_session").ok()).unwrap_or(0))
     }
 
     /// Deletes all conversation events for `agent_id`.
--- a/crates/orchestrator/src/websocket.rs
+++ b/crates/orchestrator/src/websocket.rs
@@ -1,7 +1,11 @@
 use crate::approvals::ApprovalRegistry;
 use crate::manager::AgentManager;
 use crate::scheduler::events::{EventBus, SystemEvent};
-use crate::types::{ActivityState, ApprovalDecision, ResultInfo, ToolPolicy, UsageSnapshot};
+use crate::storage::AgentStorage;
+use crate::types::{
+    ActivityState, ApprovalDecision, ConversationEvent, ConversationEventType, ResultInfo,
+    ToolPolicy, UsageSnapshot,
+};
 use axum::{
     extract::{
         ws::{Message, WebSocket},
@@ -46,6 +50,10 @@ pub struct ConnectionRegistry {
     pub approvals: ApprovalRegistry,
     /// Optional event bus for publishing lifecycle events.
     event_bus: Option<Arc<EventBus>>,
+    /// Optional storage backend for persisting conversation events.
+    storage: Option<AgentStorage>,
+    /// Per-agent session counter — incremented on every context clear.
+    session_numbers: Arc<RwLock<HashMap<Uuid, i64>>>,
 }
 
 impl Default for ConnectionRegistry {
@@ -66,6 +74,8 @@ impl ConnectionRegistry {
             connect_notify: Arc::new(tokio::sync::Notify::new()),
             approvals: ApprovalRegistry::new(300), // 5-minute default timeout
             event_bus: None,
+            storage: None,
+            session_numbers: Arc::new(RwLock::new(HashMap::new())),
         }
     }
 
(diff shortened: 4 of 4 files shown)


ShawnSunClio (changes requested):
## Review — feat(orchestrator): persist agent stream events in WebSocket handler

**Stack note:** This PR bases on `issue-1159` (PR #1167, still open) — sibling of #1168 and #1170. All three need #1167 to merge first. Diff is conflict-free relative to the parent.

---

### Blocking issues (2)

#### 1. Race condition at reconnect: events get session_number=0 during async lookup window

`register()` sets the in-memory session counter to `0` immediately, then spawns an async task to read the real value from storage:

```rust
// Sets to 0 right now
self.session_numbers.write().await.entry(agent_id).or_insert(0);

// Correct value arrives asynchronously — could be seconds later under load
tokio::spawn(async move {
    match storage.get_usage_stats(&agent_id).await {
        Ok(stats) => {
            session_numbers.write().await.insert(agent_id, stats.session_count as i64);
        }
        ...
    }
});
```

Any `assistant`, `result`, or user-message event that arrives before the async task completes will be tagged with `session_number: 0`, regardless of the agent's actual session history. For a reconnecting agent (restart, network blip, crash-recovery) that has already had two or three sessions, every event during this window lands in session 0 — permanently misattributed with no way to correct it after the fact.

The window is typically short, but during Sidekiq load or SQLite contention it could be seconds. Reconnection is the exact scenario where correct session attribution matters most (distinguishing the new run from prior ones).

**Options:**
- Read the session number synchronously before completing registration (accept the small latency hit at connect time)
- Defer event persistence until the session counter is confirmed (buffer events in a `VecDeque`, flush when the lookup resolves)
- Query the max `session_number` from `conversation_events` directly in the async task (no round-trip through `agent_sessions`) as a faster path

---

#### 2. No tests for any of the new persistence logic

The PR adds:
- `persist_event()` — fire-and-forget spawn, error logging
- `get_session_number()` — async RwLock read
- `persist_context_cleared()` — counter write + persist + stream broadcast
- `register()` session init + async lookup race
- `unregister()` session cleanup
- `send_user_message()` two new persist calls
- `handle_incoming_message()` — six new `persist_event` call sites across three match arms

None of these paths have tests. The existing websocket test suite (`mod tests` at line 1227) covers connection lifecycle and message routing but has no coverage for the persistence layer.

At minimum the following should be tested before merge:
- `register()` initialises session counter to 0 when storage is absent
- `unregister()` removes the session counter entry
- `persist_context_cleared()` updates the counter and calls `persist_event`
- `send_user_message()` calls `persist_event` twice (prompt_sent + activity_changed)
- An event persisted in `handle_incoming_message` for each of the three `assistant`/`result` arms

These are all testable with an in-memory SQLite storage instance (the same pattern already used in `storage.rs` tests via `create_test_storage()`).

---

### Non-blocking observations

#### 3. `ActivityChanged` events double the per-agent event volume

Every interaction cycle writes 2 `ActivityChanged` events (busy on `send_user_message`, idle on `result`) on top of the content events. For an active agent producing 20 assistant messages per hour, `ActivityChanged` pairs alone add ~40 rows/hour — roughly 40% of event volume for a typical run. Given the 50 k default cap from PR #1168, this reduces effective conversation history retention. Consider whether `ActivityChanged` events need to be persisted at all (vs. just broadcast on the stream), or document the trade-off explicitly.

#### 4. Session semantics at initial connect rely on agent_sessions row count

`register()` initialises the counter to `stats.session_count`, assuming that the count of session rows equals the current session number. This holds as long as the first session is numbered 0 and `clear_context()` is the only thing that creates new session rows. If that invariant ever changes (e.g., seeding, import, a new workflow phase that rotates sessions), the session tagging silently diverges. A more robust approach would be to read `MAX(session_number)` directly from `conversation_events` for the agent, which is always self-consistent.

#### 5. Minor: `(*storage).clone()` in `main.rs`

`AgentStorage` derives `Clone` and wraps a `DatabaseConnection` (SeaORM's Arc-backed pool). The clone shares the underlying connection pool, which is correct. The `(*storage)` deref is needed because `storage` is `Arc<AgentStorage>` — this is fine, just worth a comment that it's intentionally cloning the pool handle, not duplicating the DB connection.

geoffjay (commented):
## Review: feat(orchestrator): persist agent stream events in WebSocket handler

**Stack note:** This PR is stacked on `issue-1159` (PR #1167, `needs-rework`). It cannot merge until #1167 is fixed. The code review below is independent of that dependency.

---

### 🔴 Fix required — session number initialization is off-by-one

**`websocket.rs`, `register()`:**

```rust
match storage.get_usage_stats(&agent_id).await {
    Ok(stats) => {
        session_numbers.write().await.insert(agent_id, stats.session_count as i64);
    }
```

`AgentUsageStats::session_count` is documented as "Total number of sessions (including the current one, if any)" — confirmed by the storage tests:

```rust
// After start_new_session:
assert_eq!(stats.session_count, 2);  // session 0 ended + session 1 active = 2

// After recording usage (no explicit end):
assert_eq!(stats.session_count, 1);  // session 0 active = 1
```

The invariant holds: `session_count = session_number + 1` whenever a current usage session is active. Any agent that has processed at least one message will have an active session, so on reconnect:

| Context clears | Correct `session_number` | `stats.session_count` | Used value |
|---|---|---|---|
| 0 (but has usage) | 0 | 1 | **1 ← wrong** |
| 1 | 1 | 2 | **2 ← wrong** |
| N | N | N+1 | **N+1 ← wrong** |

The only case that accidentally works is a brand-new agent with no recorded usage at all (`session_count = 0 = session_number`).

**Consequence:** On reconnect, events in the current session get tagged as `session_number = N+1` instead of `N`. The REST API (#1161) and UI (#1162) would show them as belonging to a non-existent session.

**Fix — add a storage method that queries `conversation_events` directly:**

```rust
// storage.rs — new method on AgentStorage
pub async fn get_max_conversation_session_number(&self, agent_id: Uuid) -> Result<i64> {
    use sea_orm::{ConnectionTrait, Statement, DbBackend};
    let row = self.db.query_one(Statement::from_sql_and_values(
        DbBackend::Sqlite,
        "SELECT COALESCE(MAX(session_number), 0) FROM conversation_events WHERE agent_id = ?",
        [agent_id.to_string().into()],
    )).await?;
    Ok(row.and_then(|r| r.try_get_by_index::<i64>(0).ok()).unwrap_or(0))
}
```

Then in `register()`:

```rust
tokio::spawn(async move {
    match storage.get_max_conversation_session_number(agent_id).await {
        Ok(session) => {
            session_numbers.write().await.insert(agent_id, session);
        }
        Err(e) => {
            warn!(%agent_id, error = %e,
                "Failed to initialize session number from storage; defaulting to 0");
        }
    }
});
```

This is always correct because it reads directly from the same table where session numbers are written — no dependency on usage session semantics.

Alternatively (simpler, no new method), adjust the current approach:

```rust
// session_count includes the current active session, so subtract 1 when active
let session_number = if stats.current_session.is_some() {
    (stats.session_count as i64).saturating_sub(1)
} else {
    stats.session_count as i64
};
session_numbers.write().await.insert(agent_id, session_number);
```

The `get_max_conversation_session_number` approach is preferred — it doesn't rely on the relationship between usage sessions and conversation sessions staying in sync.

---

### ✅ What's correct — solid implementation throughout

**Fire-and-forget pattern** (`persist_event`): spawns a task, logs `warn` on error, no-op when `storage` is None — exactly right. ✅

**Session counter lifecycle:**
- `new()`: initialised to empty `HashMap` ✅
- `register()`: `or_insert(0)` provides an immediate safe default ✅
- `unregister()`: `remove(agent_id)` prevents unbounded growth ✅
- `persist_context_cleared()`: `insert(agent_id, new_session_number)` updates the counter before persisting the event ✅

**`handle_incoming_message()` — session read strategy:** The single `let session = registry.get_session_number(agent_id).await` at the top of the NDJSON loop means all events from one message batch share one session number and one lock acquisition. ✅

**Event coverage** vs acceptance criteria in #1160:
- `output` (per non-empty line) ✅
- `tool_use` (with full `tool_use` JSON as metadata) ✅
- `thinking` ✅
- `result` (with usage metadata including `is_error`) ✅
- `prompt_sent` + `activity_changed(busy)` from `send_user_message` ✅
- `context_cleared` via `persist_context_cleared` ✅
- `activity_changed(idle)` from result arm ✅

**Fixed `session_number: 0` hardcode** in the `agent:usage_update` stream event (flagged in pre-implementation analysis) ✅

**`main.rs`**: `(*storage).clone()` correctly dereferences the `Arc<AgentStorage>` before cloning the inner value (which is `Clone` via `DatabaseConnection`). ✅

**`manager.rs`**: `persist_context_cleared` called after `storage.end_session()` completes and before the API returns — no race possible since the user cannot send a message while `clear_context` is in-flight. ✅

**New `agent:context_cleared` stream event** in `persist_context_cleared`: additive change, existing UI clients ignore unknown event types safely. ✅

geoffjay:

- #1167
    - #1172 ◀

PR #1191, opened 2026-05-06
feat(github): add assignee/assignees filter fields to GitHub triggers

### LINKED ISSUE #1183: Add assignee/assignees filter fields to GitHub triggers
## Summary

Add optional `assignee` and `assignees` filter fields to the existing GitHub trigger source types (`github_issues` and `github_pull_requests`). This brings GitHub triggers to parity with the new GitLab triggers and the existing Linear `assignee` filter.

## Context

The issue description for #1178 explicitly notes:
"Note that assignee was added to the issues trigger and assignees to the merge requests, this is an optional field that I want to add, and I've actually been meaning to add to GitHub as well."

Currently, GitHub triggers fetch all matching issues/PRs by labels and state but have no assignee filtering. The `gh` CLI supports `--assignee` for filtering.

## Acceptance Criteria

- [ ] Add `assignee: Option<String>` to `TriggerConfig::GithubIssues` in `crates/orchestrator/src/scheduler/types.rs`
- [ ] Add `assignees: Option<Vec<String>>` to `TriggerConfig::GithubPullRequests` in `crates/orchestrator/src/scheduler/types.rs`
- [ ] Both fields use `#[serde(default)]` for backward compatibility
- [ ] Update `GithubIssueSource::new()` to accept optional `assignee` parameter
- [ ] Update `GithubIssueSource::fetch_tasks()` in `crates/orchestrator/src/scheduler/github.rs`:
  - When `assignee` is set, add `--assignee {value}` to `gh issue list` args
- [ ] Update `GithubPullRequestSource::new()` to accept optional `assignees` parameter
- [ ] Update `GithubPullRequestSource::fetch_tasks()`:
  - When `assignees` is set, add `--assignee {value}` for each assignee (gh CLI supports repeated `--assignee`)
- [ ] Update `create_source()` in `crates/orchestrator/src/scheduler/runner.rs` to pass new fields through
- [ ] Update `SourceTemplate::GithubIssues` and `SourceTemplate::GithubPullRequests` in `crates/cli/src/commands/apply.rs` with the new fields and conversion
- [ ] Update `source_github_issues` and `source_github_pull_requests` in `schemas/workflow.yml` with new optional properties
- [ ] Update existing tests in `github.rs` to verify assignee field handling
- [ ] Add test for assignee filtering in `gh` CLI args construction

## YAML Example

```yaml
# GitHub Issues with assignee filter
source:
  type: github_issues
  owner: myorg
  repo: myrepo
  labels: [agent]
  assignee: alice
  state: open

# GitHub PRs with assignees filter
source:
  type: github_pull_requests
  owner: myorg
  repo: myrepo
  assignees: [bob, carol]
  state: open
```

## Key Files

- `crates/orchestrator/src/scheduler/types.rs` - add fields to enum variants
- `crates/orchestrator/src/scheduler/github.rs` - implement filtering
- `crates/orchestrator/src/scheduler/runner.rs` - pass fields to constructors
- `crates/cli/src/commands/apply.rs` - SourceTemplate updates
- `schemas/workflow.yml` - schema updates

## Stack Base
Parallel: no ordering constraint (independent of GitLab work, touches only GitHub code)

geoffjay:
Add optional `assignee` and `assignees` filter fields to GitHub trigger sources, bringing them to parity with the Linear and GitLab triggers.

- `TriggerConfig::GithubIssues`: new `assignee: Option<String>` field with `#[serde(default)]`
- `TriggerConfig::GithubPullRequests`: new `assignees: Option<Vec<String>>` field with `#[serde(default)]`
- Extracts `build_issue_args` / `build_pr_args` helpers so arg construction is unit-testable
- `gh issue list` / `gh pr list` receive `--assignee` flag(s) when filters are set
- `SourceTemplate` in `apply.rs` updated with new fields and conversion
- CLI `display_workflow` shows assignee(s) when present
- `schemas/workflow.yml` updated with new optional properties and examples
- 5 new unit tests verify assignee arg construction with/without filters

Closes #1183

### DIFF
--- a/crates/orchestrator/src/scheduler/github.rs
+++ b/crates/orchestrator/src/scheduler/github.rs
@@ -35,11 +35,18 @@ pub struct GithubIssueSource {
     repo: String,
     labels: Vec<String>,
     state: String,
+    assignee: Option<String>,
 }
 
 impl GithubIssueSource {
-    pub fn new(owner: String, repo: String, labels: Vec<String>, state: String) -> Self {
-        Self { owner, repo, labels, state }
+    pub fn new(
+        owner: String,
+        repo: String,
+        labels: Vec<String>,
+        state: String,
+        assignee: Option<String>,
+    ) -> Self {
+        Self { owner, repo, labels, state, assignee }
     }
 }
 
@@ -66,22 +73,13 @@ struct GhAssignee {
 #[async_trait]
 impl TaskSource for GithubIssueSource {
     async fn fetch_tasks(&self) -> anyhow::Result<Vec<Task>> {
-        let mut args = vec![
-            GH_PATH.clone(),
-            "issue".to_string(),
-            "list".to_string(),
-            "--json".to_string(),
-            "number,title,body,url,labels,assignees".to_string(),
-            "--repo".to_string(),
-            format!("{}/{}", self.owner, self.repo),
-            "--state".to_string(),
-            self.state.clone(),
-        ];
-
-        for label in &self.labels {
-            args.push("--label".to_string());
-            args.push(label.clone());
-        }
+        let args = build_issue_args(
+            &self.owner,
+            &self.repo,
+            &self.labels,
+            &self.state,
+            self.assignee.as_deref(),
+        );
 
         debug!(repo = %format!("{}/{}", self.owner, self.repo), "Fetching GitHub issues");
 
@@ -123,11 +121,18 @@ pub struct GithubPullRequestSource {
     repo: String,
     labels: Vec<String>,
     state: String,
+    assignees: Option<Vec<String>>,
 }
 
 impl GithubPullRequestSource {
-    pub fn new(owner: String, repo: String, labels: Vec<String>, state: String) -> Self {
-        Self { owner, repo, labels, state }
+    pub fn new(
+        owner: String,
+        repo: String,
+        labels: Vec<String>,
+        state: String,
+        assignees: Option<Vec<String>>,
+    ) -> Self {
+        Self { owner, repo, labels, state, assignees }
     }
 }
 
@@ -150,22 +155,13 @@ struct GhPullRequest {
 #[async_trait]
 impl TaskSource for GithubPullRequestSource {
     async fn fetch_tasks(&self) -> anyhow::Result<Vec<Task>> {
-        let mut args = vec![
-            GH_PATH.clone(),
-            "pr".to_string(),
-            "list".to_string(),
-            "--json".to_string(),
-            "number,title,body,url,labels,assignees,headRefName,baseRefName,isDraft".to_string(),
-            "--repo".to_string(),
-            format!("{}/{}", self.owner, self.repo),
-            "--state".to_string(),
-            self.state.clone(),
-        ];
-
-        for label in &self.labels {
-            args.push("--label".to_string());
-            args.push(label.clone());
-        }
+        let args = build_pr_args(
+            &self.owner,
+            &self.repo,
+            &self.labels,
+            &self.state,
+            self.assignees.as_deref(),
+        );
 
         debug!(repo = %format!("{}/{}", self.owner, self.repo), "Fetching GitHub pull requests");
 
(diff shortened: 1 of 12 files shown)


geoffjay (commented):
## Review: feat(github) — assignee/assignees filter fields

**Stack note:** PR #1191 (`issue-1183`) is stacked on PR #1190 (`issue-1180`, currently `needs-rework`). The parent's rework (renaming `assignees` → `assignee` on the GitLab MR trigger) does not touch any code in this PR, so there will be no conflicts — but after #1190 is fixed and merged, a `git-spice upstack restack` will be needed before this PR can land.

---

### 1 blocking issue

#### 🔴 `assignees: Option<Vec<String>>` generates multiple `--assignee` flags that `gh pr list` does not support

`build_pr_args` emits one `--assignee <user>` flag per entry in the vec:

```rust
// github.rs — build_pr_args
if let Some(assignee_list) = assignees {
    for a in assignee_list {
        args.push("--assignee".to_string());
        args.push(a.clone());
    }
}
```

But `gh pr list` (and `gh issue list`) define the flag as a **single string**:

```
-a, --assignee string   Filter by assignee
```

Because it is a `StringVarP` (not `StringArrayVar`) in cobra/pflag, each repeated `--assignee` overwrites the previous value — only the last entry in the vec is actually used for filtering, with no error and no warning. A user who configures:

```yaml
assignees:
  - alice
  - bob
```

will get only `bob`'s PRs. `alice` is silently discarded. This is the same silent-truncation class of bug flagged on the GitLab MR trigger in PR #1190.

The test `test_build_pr_args_with_multiple_assignees` confirms and enshrines the broken behaviour — it asserts that 2 `--assignee` flags are emitted, which is exactly what causes the problem at runtime.

**Fix** — rename `assignees: Option<Vec<String>>` → `assignee: Option<String>` everywhere, to match the actual `gh pr list` flag contract and be consistent with the issue-trigger field of the same name:
- `TriggerConfig::GithubPullRequests` in `types.rs`
- `SourceTemplate::GithubPullRequests` in `cli/apply.rs`
- `GithubPullRequestSource` struct and its `new` constructor in `github.rs`
- `build_pr_args` signature: `assignees: Option<&[String]>` → `assignee: Option<&str>` — emit a single `--assignee` flag when `Some`
- `runner.rs` call site
- `cli/orchestrator.rs` display block and `create_workflow` stub
- Tests: replace `test_build_pr_args_with_multiple_assignees` with a single-value equivalent; update the no-assignee test naming
- `schemas/workflow.yml`: change `assignees` (array) → `assignee` (string) in the PR section, matching the issues section

Note: if multi-assignee GitHub PR filtering is a future requirement, the correct mechanism is `gh pr list --search "assignee:alice assignee:bob"`, not repeated `--assignee` flags.

---

### Non-blocking observations

**1 — Consistent extraction of `build_issue_args` / `build_pr_args` is excellent**

Moving arg construction into `pub(crate)` helpers that are directly unit-tested is the right pattern. The issue-side implementation (`assignee: Option<&str>` → single `--assignee` flag) is correct and clean.

**2 — Backward-compatible serde defaults**

The `#[serde(default)]` annotations on all new fields mean existing stored workflows deserialise without error. The integration tests are updated correctly to pass explicit `None` values, keeping the fixtures explicit.

**3 — `assignee: None` hardcoded in CLI `create_workflow` for GitHub types**

The `orchestrator create-workflow` command fills `assignee: None` / `assignees: None` unconditionally. This is an existing pattern (Linear and GitLab do the same), so it is not a regression — but it means the CLI subcommand cannot set assignee filters today; only the `apply` YAML path can. Worth a `// TODO` comment or a follow-up issue.

**4 — `is_draft` still not in `KNOWN_VARIABLES`**

The GitHub PR source inserts `is_draft` into task metadata (observed in `github.rs`), but `is_draft` is absent from `template.rs`'s `KNOWN_VARIABLES`. This pre-dates this PR and is out of scope here, but it means `{{is_draft}}` generates a spurious 'unknown variable' warning. Track for follow-up.

crates/orchestrator/src/scheduler/github.rs
```diff
@@ -208,6 +204,68 @@ impl TaskSource for GithubPullRequestSource {
...
+        format!("{}/{}", owner, repo),
+        "--state".to_string(),
+        state.to_string(),
+    ];
+    for label in labels {
+        args.push("--label".to_string());
+        args.push(label.clone());
+    }
+    if let Some(assignee_list) = assignees {
+        for a in assignee_list {
```
geoffjay: Bug: this loop emits multiple `--assignee` flags, but `gh pr list --assignee` is declared as `string` (not `stringArray`) in the gh CLI. Each repeated flag overwrites the previous value in pflag/cobra — only the last entry in `assignees` is used for filtering. Confirmed by `gh pr list --help`: `-a, --assignee string`. Rename the field and parameter to `assignee: Option<&str>` and emit at most one `--assignee` flag, matching the issue-trigger approach.

crates/orchestrator/src/scheduler/github.rs
```diff
@@ -346,4 +404,52 @@ mod tests {
...
+    #[test]
+    fn test_build_pr_args_with_multiple_assignees() {
+        let assignees = vec!["alice".to_string(), "bob".to_string()];
+        let args = build_pr_args("myorg", "myrepo", &[], "open", Some(&assignees));
+        let assignee_count = args.iter().filter(|a| a.as_str() == "--assignee").count();
+        assert_eq!(assignee_count, 2);
+        assert!(args.contains(&"alice".to_string()));
+        assert!(args.contains(&"bob".to_string()));
+    }
 }
```
geoffjay, line 455: This test asserts and cements the broken behaviour: it verifies that 2 `--assignee` flags are emitted, which is exactly what causes silent truncation at runtime. Once the field is renamed to `assignee: Option<String>`, replace this test with a single-value variant analogous to `test_build_issue_args_with_assignee`.

geoffjay (commented):
Inline annotation — see main review comment for full context.

Issue #1211, opened 2026-05-12
Implement skill materialization during agent spawn


geoffjay:
## Context

Part of #1208 (custom agent skills). Once agents can declare skills in YAML (#1210) and the orchestrator can discover them (#1209), the spawn pipeline needs to materialize those skills into the agent's runtime environment so Claude Code discovers them.

## Background

Claude Code automatically discovers skills from `.claude/skills/` in the working directory. The most reliable injection mechanism is to write the assigned skill files into that directory before launching the agent process. This preserves full skill semantics (triggers, slash-command invocation, dynamic loading).

### Agent spawn flow (current)

In `crates/orchestrator/src/manager.rs:spawn_agent()`:
1. Create DB record (status: Pending)
2. Create backend session (tmux/subprocess/docker)
3. `build_claude_command()` constructs the CLI invocation
4. `send_command()` launches claude in the session
5. Deliver initial prompt

Skill materialization slots in between steps 2 and 3.

### 1. Skill Materialization Function

Add to `crates/orchestrator/src/skills.rs`:

```rust
/// Write skill files into the agent's .claude/skills/ directory.
///
/// For each skill name in the agent's config, copies the skill content
/// from the discovered source into `<working_dir>/.claude/skills/<name>/SKILL.md`.
///
/// Creates the directory structure if it does not exist.
/// Does NOT overwrite existing skill files (agent's own skills take precedence).
pub async fn materialize_skills(
    working_dir: &Path,
    skill_names: &[String],
    discovered_skills: &[Skill],
) -> Result<MaterializeResult>

pub struct MaterializeResult {
    pub written: Vec<String>,     // skills successfully written
    pub skipped: Vec<String>,     // skills that already existed
    pub not_found: Vec<String>,   // skill names not in discovered set
}
```

### 2. Integration in spawn_agent()

In `manager.rs:spawn_agent()`, after creating the session and before building the command:

```rust
// Materialize skills into agent's working directory
if !config.skills.is_empty() {
    let discovered = skills::discover_skills(&skills_dir)?;
    let result = skills::materialize_skills(
        Path::new(&config.working_dir),
        &config.skills,
        &discovered,
    ).await?;
    
    if !result.not_found.is_empty() {
        tracing::warn!(
            agent = %agent_id,
            missing = ?result.not_found,
            "some requested skills were not found"
        );
    }
    
    tracing::info!(
        agent = %agent_id,
        written = result.written.len(),
        skipped = result.skipped.len(),
[...]

PR #1226, opened 2026-05-13
feat(orchestrator): add Skill model and discovery from .agentd/skills/

### LINKED ISSUE #1209: Define skill model and add discovery from .agentd/skills/
## Context

Part of #1208 (custom agent skills). Agents running in agentd currently have no way to receive project-level or agent-specific skills. Claude Code discovers skills from `.claude/skills/` in the working directory, but there is no agentd-managed skill layer.

This is the foundation issue: define what an agentd skill looks like and implement discovery from a `.agentd/skills/` directory.

## Background

The project already has 16 skills in `.claude/skills/` (agent-ops, git-spice, service-ops, etc.). These are Claude Code skills - markdown files with YAML frontmatter containing `name`, `description`, and optional trigger metadata. Example structure:

```
.claude/skills/agent-ops/SKILL.md
.claude/skills/git-spice/SKILL.md
.claude/skills/service-ops/SKILL.md
```

Each file has frontmatter like:
```yaml
---
name: agent-ops
description: Create, monitor, message, and debug agents via the CLI.
---

# Agent Operations
...markdown content...
```

### 1. Skill Model

Add a `Skill` struct (likely in `crates/orchestrator/src/types.rs` or a new `crates/orchestrator/src/skills.rs` module):

```rust
/// A discoverable skill that can be assigned to agents.
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct Skill {
    /// Skill name (directory name or filename stem).
    pub name: String,
    /// Human-readable description from frontmatter.
    pub description: Option<String>,
    /// Full markdown content (including frontmatter).
    pub content: String,
    /// Filesystem path where the skill was discovered.
    pub source_path: PathBuf,
}
```

### 2. Skill Discovery

Implement a function that scans a directory for skills:

```rust
/// Discover skills from a directory.
///
/// Supports two layouts (matching Claude Code conventions):
/// - `skills_dir/<name>/SKILL.md`
/// - `skills_dir/<name>.md`
pub fn discover_skills(skills_dir: &Path) -> Result<Vec<Skill>>
```

### 3. Discovery Paths

Skills should be discoverable from multiple locations with precedence:
1. `.agentd/skills/` (project-level, primary)
2. `~/.config/agentd/skills/` (user-level, fallback)

### 4. API Endpoint

Add a `GET /skills` endpoint to the orchestrator that returns discovered skills:

```json
[
  {
    "name": "git-spice",
    "description": "Branch stacking and PR management...",
    "source_path": ".agentd/skills/git-spice/SKILL.md"
  }
]
```

## Acceptance Criteria

- [ ] `Skill` struct defined with name, description, content, source_path
- [ ] `discover_skills()` function scans a directory for `SKILL.md` files
- [ ] Supports both `<name>/SKILL.md` and `<name>.md` layouts
- [ ] Parses YAML frontmatter to extract name and description
- [ ] `GET /skills` endpoint on orchestrator returns discovered skills
- [ ] Unit tests for discovery (empty dir, single skill, multiple skills, malformed files)

## Relevant Files

- `crates/orchestrator/src/skills.rs` - new module (primary deliverable)
- `crates/orchestrator/src/types.rs` - Skill struct (or in skills.rs)
- `crates/orchestrator/src/api.rs` - add GET /skills route
- `crates/orchestrator/src/lib.rs` - add `pub mod skills;`
- `.agentd/skills/` - new directory (created in a later issue)

## Stack Base

Parallel: no ordering constraint (this is the foundation issue)

## References

- Parent plan: #1208
- Existing skills in `.claude/skills/` for format reference

geoffjay:
Implements the foundation of the #1208 skills epic: skill model, frontmatter parsing, filesystem discovery, and the `GET /skills` API endpoint.

## What was added

**`crates/orchestrator/src/skills.rs`** (new module):
- `Skill` struct — `name`, `description`, `content`, `source_path`
- `parse_frontmatter()` — zero-dependency `---…---` block parser; extracts `name` and `description` fields
- `discover_skills(dir)` — scans a directory for both layouts: `<name>/SKILL.md` and `<name>.md`; results sorted by name
- `discover_all_skills()` — merges `.agentd/skills/` (project-level) and `~/.config/agentd/skills/` (user-level) with project taking precedence on name collision

**`crates/orchestrator/src/api.rs`**:
- `GET /skills` route added — returns `discover_all_skills()` as a JSON array; no auth required (read-only, local filesystem)

**`crates/orchestrator/src/lib.rs`**:
- `pub mod skills;` declaration added

## Tests (16, all passing)

Frontmatter: name+description, name-only, quoted values, no frontmatter, unclosed block, empty block.
Discovery: missing dir, empty dir, directory layout, flat layout, fallback names, non-`.md` ignored, mixed layouts sorted, malformed file skipped.

Closes #1209

### DIFF
--- a/crates/orchestrator/src/api.rs
+++ b/crates/orchestrator/src/api.rs
@@ -91,6 +91,7 @@ pub fn create_router(state: ApiState) -> Router {
         .route("/approvals/{id}", get(get_approval))
         .route("/approvals/{id}/approve", post(approve_tool))
         .route("/approvals/{id}/deny", post(deny_tool))
+        .route("/skills", get(list_skills))
         .route("/debug/agents", get(debug_agents))
         .route("/events/ask", post(ask_event_handler))
         // Project management
@@ -635,6 +636,16 @@ struct DebugSummary {
     running_but_disconnected: Vec<Uuid>,
     connected_but_not_running: Vec<Uuid>,
     active_workflows: usize,
+}
+
+/// `GET /skills` — list all discoverable skills.
+///
+/// Scans `.agentd/skills/` (project-level) and `~/.config/agentd/skills/`
+/// (user-level) and returns a JSON array of [`Skill`] objects.  The list is
+/// sorted by name.  Returns an empty array when no skills directory exists.
+async fn list_skills() -> impl IntoResponse {
+    let skills = crate::skills::discover_all_skills();
+    Json(skills)
 }
 
 async fn debug_agents(State(state): State<ApiState>) -> Result<impl IntoResponse, ApiError> {
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -29,6 +29,7 @@ pub mod manager;
 pub mod message_bridge;
 pub(crate) mod migration;
 pub mod scheduler;
+pub mod skills;
 pub mod storage;
 pub mod system_agents;
 pub mod types;
--- a/crates/orchestrator/src/skills.rs
+++ b/crates/orchestrator/src/skills.rs
@@ -64,7 +64,12 @@ pub struct Skill {
     /// Full Markdown content of the skill file, including frontmatter.
     pub content: String,
     /// Filesystem path where the skill was loaded from.
-    pub source_path: PathBuf,
+    ///
+    /// Used internally by the materialization step (#1211) to know which file
+    /// to copy into the agent's working directory.  Not included in API
+    /// responses — consumers only need the skill name and description.
+    #[serde(skip)]
+    pub(crate) source_path: PathBuf,
 }
 
 // ---------------------------------------------------------------------------
@@ -150,7 +155,9 @@ pub fn discover_skills(skills_dir: &Path) -> Result<Vec<Skill>> {
     let mut skills = Vec::new();
 
     for entry in std::fs::read_dir(skills_dir)? {
-        let entry = entry?;
+        // Skip individual entry errors (e.g. a single permission-denied inode)
+        // rather than aborting the entire scan.
+        let Ok(entry) = entry else { continue };
         let path = entry.path();
 
         if path.is_dir() {
@@ -189,6 +196,12 @@ pub fn discover_skills(skills_dir: &Path) -> Result<Vec<Skill>> {
 /// 1. `.agentd/skills/` — project-level
 /// 2. `~/.config/agentd/skills/` — user-level
 fn skill_search_dirs() -> Vec<PathBuf> {
+    // NOTE: ".agentd/skills" is resolved relative to the orchestrator's CWD.
+    // In development (started from the repo root) this works as expected.
+    // In production (systemd, Docker) the CWD is typically not the project
+    // root, so project-level skills may not be found.  A future enhancement
+    // should allow an explicit project-root config key (e.g.
+    // AGENTD_PROJECT_ROOT) to override this path.
     let mut dirs = vec![PathBuf::from(".agentd/skills")];
 
     // User-level fallback: $HOME/.config/agentd/skills
@@ -219,6 +232,9 @@ pub fn discover_all_skills() -> Vec<Skill> {
         }
     }
 
+    // Sort globally so the merged result is deterministic regardless of which
+    // names came from the project-level vs. user-level directory.
+    result.sort_by(|a, b| a.name.cmp(&b.name));
     result
 }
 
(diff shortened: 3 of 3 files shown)


geoffjay (commented):
## Review: feat(orchestrator): add Skill model and discovery from .agentd/skills/

### Stack Position

Branch `issue-1209` is based on `main` — no parent PR. This is the root of the #1208 skills epic; #1210, #1211, and #1213 all stack on this and cannot proceed until it lands.

---

### Blocking: `discover_all_skills()` Is Not Globally Sorted

The `GET /skills` handler doc promises: **"The list is sorted by name."**

`discover_skills()` sorts results within each directory, but `discover_all_skills()` concatenates those sorted subsets without a final sort:

```rust
// Current — result is sorted per-directory, not globally:
// project skills: ["beta", "zap"]
// user skills:    ["alpha", "gamma"]
// result:         ["beta", "zap", "alpha", "gamma"]  ← wrong

pub fn discover_all_skills() -> Vec<Skill> {
    ...
    for dir in skill_search_dirs() {
        if let Ok(skills) = discover_skills(&dir) {
            for skill in skills {
                if seen.insert(skill.name.clone()) {
                    result.push(skill);  // no global sort
                }
            }
        }
    }
    result  // ← missing sort
}
```

Fix: add one line before `result`:

```rust
    result.sort_by(|a, b| a.name.cmp(&b.name));
    result
```

This is a correctness bug relative to the documented API contract. The single-directory case works fine, but the multi-directory case (project + user) produces non-deterministic ordering depending on which names appear in which source.

---

### Non-Blocking: Per-Entry `read_dir` Errors Abort the Directory Scan

The doc for `discover_skills` says: *"Files that cannot be read or are otherwise invalid are silently skipped."*

That holds for `load_skill_file` errors (`if let Ok(skill) = ...`), but individual `DirEntry` failures propagate and abort the entire scan:

```rust
for entry in std::fs::read_dir(skills_dir)? {
    let entry = entry?;   // ← aborts the loop on any entry error
```

A single permission-denied entry mid-scan would cause `discover_skills` to return `Err`, which `discover_all_skills` then silently drops — so the whole directory is skipped rather than just the bad entry. Consistent fix:

```rust
    let Ok(entry) = entry else { continue; };
```

In practice this is extremely rare (directory entry reads rarely fail after `read_dir` succeeds), but it aligns the implementation with the stated contract.

---

### Non-Blocking: `source_path` Exposes Server Filesystem Paths via API

`Skill.source_path` serializes to an absolute path string in the `GET /skills` JSON response (e.g. `"/home/user/project/.agentd/skills/git-spice/SKILL.md"`). API consumers — including the CLI in #1213 — don't need to know where on the server's filesystem a skill lives. Consider:

```rust
// Option A — skip in serialization (simplest):
#[serde(skip)]
pub source_path: PathBuf,

// Option B — keep source_path for internal use, add a separate API response type
```

`source_path` is genuinely useful internally (for #1211 materialization — knowing which file to copy). Keeping it `pub(crate)` with `#[serde(skip)]` preserves that without leaking it.

---

### Non-Blocking: Relative CWD Path for Project Skills

```rust
let mut dirs = vec![PathBuf::from(".agentd/skills")];
```

This resolves relative to the orchestrator's current working directory at runtime. In development (started from the repo root) this works perfectly. In production (systemd service, Docker, etc.) the CWD is often `/`, `/var/lib/...`, or wherever the init system drops the process — meaning project skills would silently not be discovered.

A comment documenting this assumption (or a future config key for the project root) would prevent a hard-to-debug production surprise.

---

### Non-Blocking: No Intra-Directory Deduplication

If both `foo/SKILL.md` and `foo.md` exist in the same directory and their frontmatter both resolve to `name: foo`, `discover_skills` will return two `Skill` structs with the same name. The dedup in `discover_all_skills` only operates across source directories, not within one. Rare edge case, but worth a note or an `is_none()` check after the `find("\n---")`.

---

### What's Working Well

- **Frontmatter parser is minimal and correct** for the expected input. `strip_prefix("---")` + `find("\n---")` correctly handles the open/close delimiters; `v.trim().trim_matches('"')` correctly strips both quoted and unquoted values including CRLF (`trim()` handles `\r`).
- **`discover_all_skills` precedence model** is right: `HashSet<String>` insertion-order dedup ensures project-level wins, and per-directory errors are swallowed so a missing user dir never blocks project skills.
- **Test coverage is solid**: 16 tests exercise all documented paths — both layouts, fallback names, non-`.md` filtering, empty/missing dirs, mixed-and-sorted output.
- **`tempfile` correctly in `[dev-dependencies]`** — not shipped in production binary.
- **`GET /skills` returns `[]` (not 500) when no skills exist** — consistent with `discover_all_skills` never erroring.
- **Module is `pub mod skills`** — correctly exposed for downstream crates (#1210, #1211) to import `discover_skills` / `Skill`.

---

### Summary

One blocking issue: `discover_all_skills()` doesn't do a final sort, violating the `GET /skills` "sorted by name" guarantee when both project and user skill directories are populated. Fix is a single `result.sort_by(...)` call. The remaining notes are non-blocking. Once fixed this is ready to land and unblock the three downstream issues.

geoffjay (commented):
## Re-Review: feat(orchestrator): add Skill model and discovery from .agentd/skills/

### Blocking Issues — Both Resolved ✓

**1. Global sort in `discover_all_skills()`** — Fixed.

```rust
// Sort globally so the merged result is deterministic regardless of which
// names came from the project-level vs. user-level directory.
result.sort_by(|a, b| a.name.cmp(&b.name));
result
```

The explanatory comment is a nice addition. The new test `test_discover_all_skills_global_sort_across_directories` uses exactly the scenario from the original review (project: `["beta", "zap"]`, user: `["alpha", "gamma"]` → merged: `["alpha", "beta", "gamma", "zap"]`) — this would have caught the original bug.

**2. Per-entry `read_dir` error handling** — Fixed.

```rust
let Ok(entry) = entry else { continue };
```

Matches the exact fix requested, with a comment explaining intent. ✓

---

### Non-Blocking Items — Also Addressed

**`source_path` API leak** — Fixed proactively. `source_path` is now `pub(crate)` with `#[serde(skip)]`, and the new `test_source_path_not_included_in_json` test verifies it doesn't appear in serialized output. ✓

**CWD assumption in production** — Documented. The comment in `skill_search_dirs()` accurately describes the production CWD concern and names `AGENTD_PROJECT_ROOT` as a future config key. ✓

---

### Summary

All issues from the initial review are resolved. The code is correct, the test suite is solid (16 tests covering all documented paths), and the fixes are clean. Adding to merge queue — this unblocks #1228 and #1229.

PR #1229, opened 2026-05-13
feat(orchestrator): implement skill materialization during agent spawn

### LINKED ISSUE #1211: Implement skill materialization during agent spawn

geoffjay:
Adds `materialize_skills()` to write skill files into an agent's `.claude/skills/` directory before the Claude process is launched.

**Changes:**
- `MaterializeResult` struct tracks which skills were written, skipped (existing file), or not found
- `materialize_skills()` writes `<working_dir>/.claude/skills/<name>/SKILL.md` for each assigned skill
- Agent-local skill files take precedence — existing files are skipped, not overwritten
- Integrated into `spawn_agent()` between session creation and `build_claude_command()`
- Missing skills produce a `warn!` log, not a hard failure — agent still launches

**Worktree approach:** Skills are written to the source `working_dir` pre-launch. Because `.claude/` is typically in `.gitignore`, worktree agents discover skills via the parent directory already wired through `additional_dirs`.

Closes #1211

### DIFF
--- a/crates/orchestrator/src/manager.rs
+++ b/crates/orchestrator/src/manager.rs
@@ -134,6 +134,53 @@ impl AgentManager {
             }
         }
 
+        // Materialize skills into the agent's .claude/skills/ directory before
+        // the claude process is launched so that Claude Code discovers them on
+        // startup.  Missing skills produce a warning rather than a hard failure
+        // so that a typo in the skills list does not prevent the agent from
+        // starting entirely.
+        if !agent.config.skills.is_empty() {
+            let discovered = crate::skills::discover_all_skills();
+            match crate::skills::materialize_skills(
+                std::path::Path::new(&agent.config.working_dir),
+                &agent.config.skills,
+                &discovered,
+            )
+            .await
+            {
+                Ok(result) => {
+                    if !result.written.is_empty() {
+                        info!(
+                            agent_id = %agent.id,
+                            written = result.written.len(),
+                            "materialized skills into .claude/skills/"
+                        );
+                    }
+                    if !result.skipped.is_empty() {
+                        info!(
+                            agent_id = %agent.id,
+                            skipped = result.skipped.len(),
+                            "skipped skill materialization (agent-local files take precedence)"
+                        );
+                    }
+                    if !result.not_found.is_empty() {
+                        warn!(
+                            agent_id = %agent.id,
+                            missing = ?result.not_found,
+                            "some requested skills were not found in discovered skill set"
+                        );
+                    }
+                }
+                Err(e) => {
+                    warn!(
+                        agent_id = %agent.id,
+                        error = %e,
+                        "skill materialization failed; agent will start without skills"
+                    );
+                }
+            }
+        }
+
         // Subprocess stdio mode takes precedence: agents communicate via
         // stdin/stdout NDJSON rather than WebSocket.
         let use_stdio = self.backend.supports_subprocess_stdio();
--- a/crates/orchestrator/src/skills.rs
+++ b/crates/orchestrator/src/skills.rs
@@ -45,6 +45,98 @@ use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 
 // ---------------------------------------------------------------------------
+// Materialization
+// ---------------------------------------------------------------------------
+
+/// Result of a [`materialize_skills`] call.
+#[derive(Debug, Default, PartialEq)]
+pub struct MaterializeResult {
+    /// Skill names whose `SKILL.md` file was successfully written.
+    pub written: Vec<String>,
+    /// Skill names that were skipped because the target file already existed.
+    ///
+    /// The agent's own `.claude/skills/<name>/SKILL.md` takes precedence over
+    /// the agentd-managed copy.
+    pub skipped: Vec<String>,
+    /// Skill names that were requested but are not in `discovered_skills`.
+    pub not_found: Vec<String>,
+}
+
+/// Write skill files into the agent's `.claude/skills/` directory.
+///
+/// For each skill name in `skill_names`, copies the skill content from
+/// `discovered_skills` into `<working_dir>/.claude/skills/<name>/SKILL.md`.
+///
+/// - Creates the directory structure if it does not exist.
+/// - Does **not** overwrite existing skill files; agent-local skills take
+///   precedence (reported as [`MaterializeResult::skipped`]).
+/// - Skill names not present in `discovered_skills` are reported in
+///   [`MaterializeResult::not_found`] rather than returning an error.
+///
+/// # Worktree agents
+///
+/// When an agent uses `--worktree`, Claude Code creates a temporary git
+/// worktree.  Skills are written to the source `working_dir` *before* launch.
+/// Because `.claude/` is typically in `.gitignore`, the worktree does not
+/// inherit those files — the agent's `additional_dirs` (already wired up in
+/// `build_claude_command`) point back at the project root where the skills live.
+pub async fn materialize_skills(
+    working_dir: &Path,
+    skill_names: &[String],
+    discovered_skills: &[Skill],
+) -> Result<MaterializeResult> {
+    use std::collections::HashMap;
+
+    let index: HashMap<&str, &Skill> =
+        discovered_skills.iter().map(|s| (s.name.as_str(), s)).collect();
+
+    let mut result = MaterializeResult::default();
+
+    for name in skill_names {
+        match index.get(name.as_str()) {
+            None => {
+                result.not_found.push(name.clone());
+            }
+            Some(skill) => {
+                let target_dir = working_dir.join(".claude").join("skills").join(name);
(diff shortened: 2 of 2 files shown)


geoffjay (commented):
## Review: feat(orchestrator): implement skill materialization during agent spawn

### Stack Position

Branch `issue-1211` is stacked on `issue-1210` (PR #1228), which is itself stacked on `epic-1209` (PR #1226). Full chain: `main ← epic-1209 ← issue-1210 ← issue-1211`.

**This PR cannot merge until #1226 and #1228 both clear.** Status at review time:
- PR #1226: re-submitted for re-review (`review-agent` label set) — blocking
- PR #1228: reviewed with comment, `merge-queue` held pending #1226

The code in this PR is reviewed in isolation below.

---

### Non-Blocking: `target_file.exists()` Is Synchronous in an Async Function

`materialize_skills` is `async` and uses `tokio::fs` for all other I/O, but the "skip if already exists" check uses the blocking `std::path::Path::exists()`:

```rust
// skills.rs ~line 104 — synchronous stat call in async context:
if target_file.exists() {
    result.skipped.push(name.clone());
    continue;
}

// vs. the async calls immediately below:
tokio::fs::create_dir_all(&target_dir).await?;
tokio::fs::write(&target_file, &skill.content).await?;
```

For the typical `.agentd/skills/` directory this is negligible, but it's an inconsistency that could mask problems on slow filesystems or network mounts. Consistent fix:

```rust
if tokio::fs::try_exists(&target_file).await.unwrap_or(false) {
    result.skipped.push(name.clone());
    continue;
}
```

`try_exists` is preferred over `metadata().is_ok()` because it returns `Ok(false)` for a missing file rather than `Err`, and the `.unwrap_or(false)` correctly falls through to a write attempt if the stat itself fails.

---

### Non-Blocking: I/O Write Errors Land in `not_found`

`MaterializeResult` has three buckets: `written`, `skipped`, `not_found`. In `materialize_skills`, the not-found path is correct — skill names with no matching entry in `discovered_skills` go to `not_found`. But the `?` propagation on `create_dir_all` and `write` would bubble the error up and abort the entire call, not route to `not_found`.

More importantly, if a future refactor changes the write errors to be swallowed and categorized, they'd naturally fall into `not_found` — but a skill that was found and failed to write is categorically different from a skill that was never discovered. A `failed` field would make this distinction explicit and is cheap to add:

```rust
#[derive(Debug, Default, PartialEq)]
pub struct MaterializeResult {
    pub written: Vec<String>,
    pub skipped: Vec<String>,
    pub not_found: Vec<String>,
    pub failed: Vec<String>,   // found but couldn't be written
}
```

The integration in `manager.rs` already logs `not_found` names as `"skills not found"` — having a separate `failed` vector would give operators the right signal when skill files exist but can't be materialized (permissions, disk full, etc.).

---

### Non-Blocking: `discover_all_skills()` Is Synchronous Inside Async `spawn_agent`

`spawn_agent` in `manager.rs` is async, but `discover_all_skills()` uses `std::fs::read_dir` throughout. The call is:

```rust
let discovered = crate::skills::discover_all_skills();  // blocking sync
match crate::skills::materialize_skills(...).await {
```

Same observation as PR #1226 review — for the typical small skills directory this is fine in practice. Worth noting as a future optimization (pass discovered skills in from the caller, or spawn via `tokio::task::spawn_blocking`). Not worth blocking the PR on.

---

### What's Working Well

- **Soft failure is correct**: `Err(e) => { warn!(...); }` — agent still starts even if skill materialization fails. This is the right tradeoff; a missing skill should not prevent an agent from launching.
- **Guard avoids unnecessary work**: `if !agent.config.skills.is_empty()` skips both discovery and materialization when no skills are requested.
- **Placement in `spawn_agent`**: materialization after `create_session` but before `build_claude_command` is correct — the working directory exists at that point, and skills must be present before the subprocess starts.
- **Target path convention**: `.claude/skills/<name>/SKILL.md` is consistent with the Claude Code skill directory layout.
- **Tracing fields**: `%result.written.len()`, `%result.skipped.len()`, `%result.not_found.len()` use `%` (Display) for counts and `?` (Debug) for the string vecs — correct.
- **`HashMap` index in `materialize_skills`**: building a `name → &Skill` map before iterating `skill_names` is O(n) rather than O(n²) — correct choice.
- **`MaterializeResult` derives**: `Default` + `PartialEq` are the right set; no `Clone` needed since the caller doesn't retain it.

---

### Summary

Two non-blocking observations: the `target_file.exists()` sync/async inconsistency (swap for `tokio::fs::try_exists`) and the missing `failed` bucket in `MaterializeResult` for write errors distinct from discovery failures. The synchronous `discover_all_skills()` in an async context is a pre-existing pattern noted in #1226 and tracked separately.

The integration is correct and the soft-failure behavior is the right call. This PR is waiting on #1226 → #1228 to clear before it can merge; no action needed here beyond that.

PR #1248, opened 2026-05-25
fix(scheduler): restart workflow runners when an agent is restarted

### LINKED ISSUE #1103: Restarting a failed agent should re-launch its associated workflow runners
## Problem

When an agent is restarted via `POST /agents/{id}/restart` (or the UI restart button), its workflow runners are not restarted. The agent comes back to `Running` status but its workflows remain dead -- no polling, no dispatches.

### Root cause

The scheduler starts workflow runners at orchestrator startup. Each runner is a tokio task that polls a task source (Linear, GitHub, cron) on an interval. When an agent fails:

1. The workflow runner waits up to 60s for the agent to reconnect
2. If the agent doesn't reconnect, the runner logs "Agent did not connect within 60s, workflow not resumed" and **exits**
3. The runner task is never respawned

When the agent is later restarted (via API, UI, or reconciliation), nothing notifies the scheduler to re-launch the dead workflow runners.

### Observed behavior

From orchestrator logs:

```
21:06:57 WARN Agent did not connect within 60s, workflow not resumed
  workflow_id=465f6a05 agent_id=3032bc51
```

After restarting the agent via `POST /agents/{id}/restart`:
- Agent status: `Running`, PID assigned, WebSocket connected
- Workflow status: `enabled=true` but **no runner task active**
- Linear polling: not happening, no dispatches created
- The only workaround is toggling the workflow (disable + re-enable) or restarting the orchestrator service

### Expected behavior

Restarting a failed agent should restore it to a fully functional state, including:
1. Agent process spawned and connected via WebSocket
2. All enabled workflows associated with that agent have active runner tasks
3. Workflow runners resume polling their task sources

### Proposed solution

When `restart_agent_by_id` (or the internal `restart_agent`) succeeds:
1. Query all enabled workflows where `agent_id` matches the restarted agent
2. For each workflow, check if a runner task is active
3. If no runner is active, start one via `scheduler.start_workflow(workflow_id)`

This could be done in the `restart_agent_by_id` method itself (requires the manager to have a reference to the scheduler), or as a reactive event -- the manager publishes an `AgentRestarted` event on the event bus, and the scheduler subscribes to re-launch runners.

The event-based approach is cleaner since it avoids coupling the manager to the scheduler.

### Files

- `crates/orchestrator/src/manager.rs` -- `restart_agent_by_id` / `restart_agent`
- `crates/orchestrator/src/scheduler/mod.rs` -- workflow runner lifecycle
- `crates/orchestrator/src/scheduler/runner.rs` -- individual runner tasks
- `crates/orchestrator/src/scheduler/events.rs` -- event bus (already has `AgentConnected`, `AgentDisconnected`)
- `crates/orchestrator/src/api.rs` -- `POST /agents/{id}/restart` handler

### Workaround

Toggle the workflow to restart its runner:
```bash
agent workflow update <workflow_id> --disable
agent workflow update <workflow_id> --enable
```

Or restart the orchestrator service.

geoffjay:
When an agent fails and its workflow runners time out waiting for it to reconnect (the 60-second startup window in `resume_workflows`), no runner is ever started for those workflows. If the agent is later restarted via the API, UI, or reconciliation, the workflows remain dead -- no polling, no dispatches.

## Changes

- **`scheduler/events.rs`**: Add `AgentRestarted { agent_id }` to `SystemEvent`. Published by `restart_agent` after the new process is launched and the DB record is updated.
- **`manager.rs`**: Publish `SystemEvent::AgentRestarted` at the end of `restart_agent` (via `self.registry.event_bus()`). This covers all restart paths: API, reconcile, bootstrap, and `clear_context`.
- **`scheduler/mod.rs`**: Add `Scheduler::restart_workflows_for_agent` which lists all enabled workflows for the agent, skips any with active runners (idempotent), and calls `start_workflow` for the dead ones.
- **`main.rs`**: Subscribe to `AgentRestarted` events in a background task that calls `restart_workflows_for_agent` reactively.
- **`config.rs`**: Fix two `OrchestratorConfig` test initializers that were missing the `backend` field, causing the test binary to fail to compile.

## How it works

1. Agent fails → 60s timeout → runner exits → `runners` map has no entry for the workflow
2. Agent is restarted via `POST /agents/{id}/restart`
3. `restart_agent` succeeds → publishes `AgentRestarted { agent_id }`
4. Event subscriber calls `restart_workflows_for_agent(agent_id)`
5. Method finds the enabled workflow with no active runner → calls `start_workflow`
6. Runner starts and begins polling its task source; dispatches once the agent connects

Closes #1103

### DIFF
--- a/crates/cli/src/commands/orchestrator.rs
+++ b/crates/cli/src/commands/orchestrator.rs
@@ -1923,8 +1923,7 @@ fn format_stream_message(text: &str, verbose: bool) {
                 if let Some(usage) = msg.get("usage") {
                     let input = usage.get("input_tokens").and_then(|v| v.as_u64()).unwrap_or(0);
                     let output = usage.get("output_tokens").and_then(|v| v.as_u64()).unwrap_or(0);
-                    let cost =
-                        usage.get("total_cost_usd").and_then(|v| v.as_f64()).unwrap_or(0.0);
+                    let cost = usage.get("total_cost_usd").and_then(|v| v.as_f64()).unwrap_or(0.0);
                     println!(
                         "{} {}",
                         prefix,
--- a/crates/mcp/src/tools/communicate.rs
+++ b/crates/mcp/src/tools/communicate.rs
@@ -201,11 +201,7 @@ pub async fn run_get_room(client: &AgentdClient, room_id: &str) -> String {
     out
 }
 
-pub async fn run_list_messages(
-    client: &AgentdClient,
-    room_id: &str,
-    limit: Option<u32>,
-) -> String {
+pub async fn run_list_messages(client: &AgentdClient, room_id: &str, limit: Option<u32>) -> String {
     let base = client.communicate_url();
     let limit_val = limit.unwrap_or(20).clamp(1, 200);
     let url = format!("{base}/rooms/{room_id}/messages?limit={limit_val}");
--- a/crates/mcp/src/tools/diagnostic.rs
+++ b/crates/mcp/src/tools/diagnostic.rs
@@ -167,7 +167,9 @@ pub async fn run_diagnose_agent(client: &AgentdClient, agent_id: &str) -> String
                 info.push("Backend is **pty** — direct PTY session.".to_string());
             }
             "tmux" => {
-                info.push("Backend is **tmux** — session can be attached for inspection.".to_string());
+                info.push(
+                    "Backend is **tmux** — session can be attached for inspection.".to_string(),
+                );
             }
             _ => {}
         }
@@ -233,10 +235,8 @@ pub async fn run_diagnose_workflow(client: &AgentdClient, workflow_id: &str) -> 
     let wf_name = str_field(&wf, "name");
     let agent_id = str_field(&wf, "agent_id");
     let enabled = wf["enabled"].as_bool().unwrap_or(false);
-    let trigger_type = wf["trigger_type"]
-        .as_str()
-        .or_else(|| wf["source"]["type"].as_str())
-        .unwrap_or("unknown");
+    let trigger_type =
+        wf["trigger_type"].as_str().or_else(|| wf["source"]["type"].as_str()).unwrap_or("unknown");
 
     writeln!(report, "# Workflow Diagnostic: {wf_name}").ok();
     writeln!(report, "- **ID:** `{workflow_id}`").ok();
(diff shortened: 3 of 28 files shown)


geoffjay (commented):
## Review: fix(scheduler): restart workflow runners when an agent is restarted

The core approach is correct and well-designed. The event-bus pattern is consistent with the existing AgentDisconnected handler in main.rs, idempotency is properly handled, and the fix targets the root cause. Two blocking issues need to be resolved before merge.

### Blocking

**1. CI Format check is failing** -- The Format CI job (cargo fmt --check) completed with FAILURE. Run cargo fmt, commit, and push before merge.

**2. Wrong base branch** -- This PR targets main, but CLAUDE.md requires all feature work to target feature/autonomous-pipeline. Please retarget: gh pr edit 1248 --repo geoffjay/agentd --base feature/autonomous-pipeline

### Non-blocking suggestions

**3. restart_workflows_for_agent receiver (scheduler/mod.rs)** -- The method takes self: &Arc<Self> but never clones or stores the Arc. Unlike resume_workflows, no background spawns are done here, so &self is sufficient and more consistent with start_workflow, stop_workflow, and other methods in this impl block.

**4. list_workflows(None) scans all workflows on every restart (scheduler/mod.rs line 417)** -- Storage only supports filtering by project_id, not agent_id. Every agent restart loads all workflows and filters in Rust. Fine for current scale since restarts are rare, but a follow-up to add agent_id filtering to SchedulerStorage::list_workflows would prevent it becoming a hotspot.

**5. Misleading error-level log for a benign race (scheduler/mod.rs ~line 449)** -- A lingering resume_workflows background task (still inside its 60-second wait) could race with restart_workflows_for_agent and call start_workflow for the same workflow. The loser emits error\!(...Failed to re-launch workflow runner...) when the root cause is just Workflow X is already running, which is benign. Consider warn\! or checking the error variant to distinguish this case from genuine failures.

### What is well done

- Complete coverage: all restart paths flow through the private restart_agent method, so the single publish site in manager.rs covers API, reconcile, bootstrap, and clear_context.
- Correct idempotency: the runners.contains_key check plus start_workflow own guard prevent double-starting runners.
- Consistent pattern: the event subscriber in main.rs is structurally identical to the existing AgentDisconnected handler.
- Test coverage: test_agent_restarted_event and the updated test_publish_all_event_variants cleanly cover the new variant; the bus capacity bump from 16 to 32 in the all-variants test is correct.
- config.rs fix: adding the missing backend field to the two test initializers is a genuine compilation fix that belongs in this PR.

Please fix the format failure and retarget the base branch, then re-submit for review.

PR #1252, opened 2026-06-07
fix(build): use rustls reqwest in all crates to fix musl release build

geoffjay:
## Root cause

The `x86_64-unknown-linux-musl` release build failed because `openssl-sys` was pulled into the dependency graph via `reqwest`'s `default-tls` → `native-tls` feature, and system OpenSSL cannot cross-compile against musl without a sysroot:

```
Could not find openssl via pkg-config: pkg-config has not been configured to support cross-compilation.
openssl-sys = 0.9.112
```

Six crates (`ask`, `cli`, `communicate`, `mcp`, `notify`, `wrap`) declared reqwest with default features. Cargo feature unification then dragged native OpenSSL into the entire workspace build, even though the workspace standardizes on rustls.

## Fix

Switch these crates to the workspace reqwest dependency (`rustls-tls`, `default-features = false`), matching the existing pattern in `core`, `index`, `hook`, `monitor`, `memory`, and `ui`. No native-tls-specific reqwest APIs were in use.

## Verification

- `cargo tree -i openssl-sys --target x86_64-unknown-linux-musl` → `openssl-sys` no longer present
- `cargo check --workspace --bins` passes
- clippy clean (pre-commit hook)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1275,16 +1275,6 @@ checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
 
 [[package]]
 name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
-name = "core-foundation"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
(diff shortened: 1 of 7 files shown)


geoffjay (commented):
## Review: LGTM ✅

**Stack position:** Branches directly off `main` — no parent PR in the stack.

### What the PR does

Switches six crates (`ask`, `cli`, `communicate`, `mcp`, `notify`, `wrap`) from an inline `reqwest = { version = "0.12", features = ["json"] }` to the workspace declaration `reqwest = { workspace = true, features = ["json"] }`. The workspace definition carries `default-features = false, features = ["rustls-tls"]`, so the effective per-crate feature set becomes `["rustls-tls", "json"]` — which is exactly what the rest of the workspace already uses.

### Correctness

Root cause analysis is accurate. Cargo feature unification means even a single crate activating `native-tls` drags `openssl-sys` into the entire workspace build graph. Removing all six offending declarations is the right fix.

### Lock file

All expected packages are gone: `openssl`, `openssl-sys`, `openssl-macros`, `native-tls`, `hyper-tls`, `tokio-native-tls`, `foreign-types`, `foreign-types-shared`, `system-configuration`, `system-configuration-sys`, `windows-registry`, and the duplicate `core-foundation 0.9.4`. The remaining `openssl-probe` is a transitive dep of `rustls-native-certs` (runtime CA bundle discovery, no native OpenSSL linking) — harmless on musl targets.

### Remaining inline declarations

`common`, `orchestrator`, and `tui` still declare reqwest inline with explicit `default-features = false, rustls-tls` — they were never broken. Normalizing them to `workspace = true` would be a nice follow-up consistency improvement but is not required here.

### Scope

Minimal and focused. No source changes, no unrelated refactoring. No blocking issues found.

> Note: Cannot self-approve — marking merge-queue directly.

PR #1256, opened 2026-06-10
feat(install): release tarballs, checksums, and curl|sh installer (phase 2)

geoffjay:
Implements Phase 2 of docs/planning/installation-method.md as a hand-rolled
release pipeline rather than cargo-dist (which models each workspace package
as its own app and cannot bundle the multi-crate binary suite into a single
installer; rationale recorded in the planning doc).

Release workflow:
- build the web UI once with bun and include it in every tarball under ui/
- package per-target tarballs (agentd-<tag>-<target>.tar.gz) with bare-named
  binaries (agent, agentd-*) so `agent install --bin-src` can discover them
- add all 10 service binaries (agentd-hook/monitor/ui were missing) and fail
  the release on a missing binary instead of silently skipping
- add aarch64-unknown-linux-musl on the ubuntu-24.04-arm runner
- generate SHA256SUMS and upload it plus the installer as release assets
- publish tags containing '-' (e.g. v0.5.0-rc.1) as pre-releases
- drop the loose <bin>-<target> assets (target-suffixed names were never
  discoverable by `agent install` and nothing consumed them)

Installer (contrib/scripts/install.sh, replaces the legacy interactive
build-from-source script): POSIX sh, detects the platform, resolves the
latest tag via the releases/latest redirect (AGENTD_VERSION pins a version),
downloads with curl or wget, verifies the tarball against SHA256SUMS,
extracts to a temp dir, and runs `agent install` so the platform/service/
migration logic stays in Rust. cargo-binstall and Homebrew are descoped.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,8 +9,31 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  build-ui:
+    name: Build UI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install bun
+        uses: oven-sh/setup-bun@v2
+
+      - name: Build UI
+        working-directory: ui
+        run: |
+          bun install --frozen-lockfile
+          bun run build
+
+      - name: Upload UI assets
+        uses: actions/upload-artifact@v4
+        with:
+          name: ui-dist
+          path: ui/dist
+          if-no-files-found: error
+
   build:
     name: Build (${{ matrix.target }})
+    needs: build-ui
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -18,6 +41,9 @@ jobs:
         include:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
+          # GitHub-hosted arm64 runners are free for public repositories only.
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-24.04-arm
           - target: x86_64-apple-darwin
             os: macos-15-intel
           - target: aarch64-apple-darwin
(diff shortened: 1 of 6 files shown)


geoffjay (commented):
## Review: LGTM

**Stack position:** `install-method-phase-2` was stacked on `feature/install-method (#1253)`, which is now **merged**. No restack needed — this branch can go directly to `main`.

---

### Release workflow (`.github/workflows/release.yml`)

**build-ui job** — Build-once, download-everywhere is the right pattern. Artifact upload/download strips executable bits; packaging into a tarball in the build job preserves modes. The comment on lines 82-86 explains this clearly for future maintainers.

**build job improvements** — All correct:
- Added `aarch64-unknown-linux-musl` on `ubuntu-24.04-arm` (native musl-gcc via `musl-tools`). The inline comment about arm64 runners being free for public repos only is the right note to leave.
- `set -euo pipefail` added — essential for a release packaging script.
- Silent-skip on missing binaries replaced with fail-fast: a missing binary now aborts the job rather than shipping an incomplete tarball. This is the correct behavior.
- All 10 service binaries listed explicitly (`agentd-hook`, `agentd-monitor`, `agentd-ui` were previously missing). The `cli -> agent` rename is handled outside the loop as a separate, readable step.
- `chmod 755` applied before tarring — necessary since artifact downloads may not preserve the executable bit.

**release job** — Checksum generation runs on the Ubuntu host with `sha256sum`, producing the expected `<hash>  <filename>` two-space format. `fail_on_unmatched_files: true` is already set. Pre-release detection via `contains(github.ref_name, '-')` is the right lightweight heuristic for semver pre-release tags.

---

### Installer script (`contrib/scripts/install.sh`)

Complete rewrite from interactive bash to non-interactive POSIX sh. Verified against the Phase 1 CLI interface:

- `agent install --bin-src $stage --ui-dir $stage/ui` maps exactly to the `bin_src` and `ui_dir` clap args in `crates/cli/src/main.rs`. Confirmed.
- `PREFIX` env var flows correctly: the script passes it through to the subprocess environment, and `run_install()` calls `agentd_install::get_prefix()` which reads `PREFIX`. Confirmed.

**Correct patterns throughout:**
- `mktemp -d "${TMPDIR:-/tmp}/agentd-install.XXXXXX"` — portable across macOS ($TMPDIR) and Linux.
- Cleanup trap — handles success, failure, and signals correctly.
- `curl -fsSLI -o /dev/null -w '%{url_effective}'` — correct way to follow the `releases/latest` redirect without an API call.
- `grep " ${asset}\$" SHA256SUMS` — the two-space `sha256sum` output format is correctly matched by the single-space pattern anchored at end-of-line (the grep matches starting at the second space before the filename).
- `shasum -a 256 -c` fallback for macOS (where `sha256sum` may not be present).
- Quarantine removal (`xattr -dr com.apple.quarantine`) is defensive and correct, with `|| true` since it is not always needed.
- `AGENTD_VERSION` with or without the `v` prefix is normalized before use.

---

### Non-blocking observations

**`resolve_version()` wget path (lines 76-78):** The `wget -q -S --max-redirect=0` technique does not work on BusyBox wget (Alpine and similar minimal images lack both `-S` and `--max-redirect`). This is a fallback path only reached when `curl` is absent — uncommon in practice — and the explicit `[ -n "$effective" ]` guard produces a clear actionable error. Worth noting as a known gap but not blocking.

**`trap 'EXIT'` portability:** POSIX defines `trap 0` rather than `trap EXIT`. In practice every `/bin/sh` in scope (bash, dash, ash, zsh) supports `EXIT` as an alias, so this works universally. Not worth changing, just noting for the record.

---

### Documentation

README.md, contrib/README.md, docs/public/install.md, and the planning doc are all updated consistently with the new install story. The "Why not cargo-dist" section in the planning doc is exactly the right rationale to record for future contributors.

> Note: Cannot self-approve — marking merge-queue directly.

PR #1271, opened 2026-06-11
feat(monitor): Prometheus query client and curated named-query API

geoffjay:
Give the previously unused monitor service its purpose as the metrics
query/analysis layer for the platform.

- prometheus.rs: minimal hand-rolled client for /api/v1/query and
  /api/v1/query_range (typed PromData vector/matrix/scalar envelope,
  thiserror PromError, 5s timeout). No new dependencies — reqwest was
  already in the tree, and two endpoints don't justify a query-client
  crate.
- queries.rs: a 14-entry curated PromQL catalog over the real agentd
  metric inventory (dispatch success rate/throughput, agent restarts,
  approvals backlog/resolution, websocket connections, per-service HTTP
  error rate and p95 latency, session cost, notification backlog,
  message drops, host saturation, service-up). The $__window token is
  the only caller-controlled text spliced into PromQL and is validated
  as ^[0-9]+[smhdw]$ — the injection guard. Counter queries use
  increase()/rate() (reset-safe); unbounded-cardinality labels
  (room_id) are aggregated or excluded.
- API: GET /queries (static catalog, always 200), GET /queries/{name}
  with window/mode/range_minutes/step_secs, GET /query raw passthrough
  (Prometheus is loopback-only). Errors: 404 unknown name listing the
  catalog, 400 bad params, 502 when Prometheus is down — distinct from
  monitor failures so callers can tell which layer broke.
- Config: prometheus_url via AGENTD_PROMETHEUS_URL (default
  http://127.0.0.1:9090, matching infra/launchd).
- MonitorClient gains list_queries()/run_query().

### DIFF
--- a/crates/monitor/src/prometheus.rs
+++ b/crates/monitor/src/prometheus.rs
@@ -1,0 +1,240 @@
+//! Minimal Prometheus HTTP API client.
+//!
+//! Hand-rolled over `reqwest` rather than pulling in a query-client crate:
+//! only two endpoints are needed (`/api/v1/query` and `/api/v1/query_range`),
+//! and the workspace convention is thin typed reqwest wrappers.
+//!
+//! The agentd observability stack runs Prometheus at `127.0.0.1:9090`
+//! (see `infra/prometheus/` and the launchd plist) scraping every service's
+//! metrics endpoint.
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+use std::time::Duration;
+
+/// Errors from the Prometheus HTTP API.
+#[derive(Debug, thiserror::Error)]
+pub enum PromError {
+    /// Prometheus could not be reached at all.
+    #[error("Prometheus unreachable at {url}: {source}")]
+    Unreachable {
+        url: String,
+        #[source]
+        source: reqwest::Error,
+    },
+    /// Prometheus answered with a non-success HTTP status.
+    #[error("Prometheus returned HTTP {status}: {body}")]
+    HttpStatus { status: u16, body: String },
+    /// Prometheus accepted the request but rejected the query.
+    #[error("Prometheus query error ({error_type}): {error}")]
+    Query { error_type: String, error: String },
+    /// The response body did not match the expected envelope.
+    #[error("Failed to parse Prometheus response: {0}")]
+    Parse(String),
+}
+
+/// One instant-vector sample: label set + (timestamp, value) pair.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct VectorSample {
+    /// Label name → value, including `__name__` when present.
+    pub metric: BTreeMap<String, String>,
+    /// `[unix_seconds, "value"]` as returned by Prometheus.
+    pub value: (f64, String),
+}
+
+/// One range-vector series: label set + list of (timestamp, value) pairs.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MatrixSeries {
+    pub metric: BTreeMap<String, String>,
+    /// `[[unix_seconds, "value"], ...]` as returned by Prometheus.
+    pub values: Vec<(f64, String)>,
+}
+
+/// Query result data, tagged by Prometheus's `resultType`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "resultType", content = "result", rename_all = "lowercase")]
+pub enum PromData {
+    Vector(Vec<VectorSample>),
+    Matrix(Vec<MatrixSeries>),
+    Scalar((f64, String)),
+}
+
+/// The standard Prometheus HTTP API response envelope.
+#[derive(Debug, Deserialize)]
+struct PromEnvelope {
+    status: String,
+    #[serde(default)]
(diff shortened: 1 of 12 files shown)


crates/monitor/src/queries.rs
```diff
@@ -0,0 +1,237 @@
...
+        name: "http-p95-latency",
+        description: "p95 HTTP request latency per service over the window (seconds)",
+        promql: r#"histogram_quantile(0.95, sum by (service, le) (rate(http_request_duration_seconds_bucket[$__window])))"#,
+        unit: "seconds",
+        default_window: "15m",
+    },
+    NamedQuery {
+        name: "session-cost",
+        description: "Claude session spend over the window (USD, gauge delta)",
+        promql: r#"delta(usage_session_cost_usd_total[$__window])"#,
```
geoffjay, line 103: **Blocking:** `delta()` is for gauges; `usage_session_cost_usd_total` ends in `_total`, which is the Prometheus counter convention. Counter resets on service restart will produce large negative values with `delta()`. The PR description states counter queries use `increase()` for reset-safety — this entry contradicts that.

```rust
// Fix:
promql: r#"increase(usage_session_cost_usd_total[$__window])"#,
```

crates/monitor/src/prometheus.rs
```diff
@@ -0,0 +1,240 @@
...
+    ) -> Result<PromData, PromError> {
+        let response = request
+            .send()
+            .await
+            .map_err(|source| PromError::Unreachable { url: url.to_string(), source })?;
+
+        let status = response.status();
+        let body = response
+            .text()
+            .await
```
geoffjay, line 134: **Blocking:** A body-read failure occurs after Prometheus was already reachable (a response with a status code was received). Mapping it to `PromError::Unreachable` produces a misleading "Prometheus unreachable at {url}" message and conflates two distinct failure modes.

Consider:
```rust
.map_err(|e| PromError::Parse(e.to_string()))?
```
or introduce a dedicated `BodyRead { url, source }` variant.

geoffjay (commented):
## Stack Note

PR #1271 is stacked on #1270 (`sa/08-architect`), which is still open and pending its own review. This PR's changes are self-contained and reviewable now, but it cannot merge until the parent lands.

---

## Blocking Issues

### 1. `session-cost` uses `delta()` on a counter (`queries.rs`)

The metric `usage_session_cost_usd_total` ends in `_total` — the Prometheus naming convention for **counters** (monotonically increasing, resets to 0 on service restart). `delta()` is designed for **gauges** and does not handle resets; it will return a large negative value after any restart.

The PR description explicitly states _"Counter queries use increase()/rate() (reset-safe)"_ but this entry contradicts that principle. The description field calling it a "gauge delta" suggests the metric semantics may have been mis-classified.

**Fix:** Replace `delta(usage_session_cost_usd_total[$__window])` with `increase(usage_session_cost_usd_total[$__window])`.

### 2. `PromError::Unreachable` misused for body-read failure (`prometheus.rs`)

In `execute()`, the second `.map_err` maps a response-body read failure to `PromError::Unreachable`. However at that point Prometheus **was** reachable — a response was received with a status code; it is the body stream that failed (e.g. connection dropped mid-transfer). The resulting error message "Prometheus unreachable at {url}" misleads operators diagnosing the failure. The `test_prometheus_down_returns_502` test asserts `.contains("unreachable")`, which would accidentally pass for this case too, masking the mismatch.

**Fix:** Map body-read errors to `PromError::Parse(e.to_string())` (closer semantically) or introduce a `BodyRead` variant.

---

## Non-blocking Suggestions

### 3. `map_prom_error` returns 400 for named-query PromQL errors

When a **named** catalog query triggers `PromError::Query` (Prometheus rejected the PromQL), `map_prom_error` returns `ApiError::BadQueryParam` -> HTTP 400. For named queries the caller did not supply the PromQL — the catalog did. HTTP 400 implies caller error; HTTP 502 or 500 would be more accurate for a catalog issue. Consider distinguishing the two call sites, or at least logging a warning when a catalog query fails with a PromQL error.

### 4. `run_query()` in `client.rs` builds query params via string interpolation

`client.rs::run_query()` manually assembles the query string with `format!` + `join("&")`. `prometheus.rs` consistently uses reqwest's `.query()` builder, which handles URL encoding automatically. Using `.query()` here would be consistent with the rest of the crate.

### 5. No tests for new `client.rs` methods

`list_queries()` and `run_query()` are untested. The existing client test suite already mocks the server with `axum::serve`, so adding a couple of cases would follow established patterns at low cost.

---

## What Works Well

- **Injection guard** (`valid_window`) is sound and well-tested: the test matrix covers floats, negatives, injection payloads, and blank inputs. Injection vector confirmed closed.
- **14-entry catalog** is well-curated: all other counter queries correctly use `increase()`/`rate()`, unbounded-cardinality labels (`room_id`) are aggregated or excluded.
- **Error taxonomy** (502 Prometheus down vs 404 unknown name vs 400 bad param) is clean and testable.
- `windowed_queries_declare_a_default_window` invariant test is excellent defensive hygiene that will catch future catalog mistakes.
- `PromData` serde adjacently-tagged deserialization correctly matches the Prometheus envelope layout.
- `range_minutes` capped at 7 days and `step_secs` floored at 1 are sensible guards against degenerate range queries.

PR #1273, opened 2026-06-11
feat(orchestrator): agentd-analyst built-in agent

geoffjay:
Fourth and final registry agent: a lazy, read-only built-in that
analyzes platform metrics and health through the monitor-backed
query_metrics MCP tool and the orchestrator's read APIs.

- Prompt carries the full named-query catalog with healthy baselines
  and what deviations mean, plus playbooks (cost spike, latency,
  failures, routine review) and the Prometheus-down fallback to direct
  scrapes. A test pins the catalog coverage.
- Policy allows the read/diagnose tool families plus exactly two
  escalation channels (create_notification, send_room_message); no
  lifecycle, creation, approval, or remediation tools.

Suggested follow-up (dogfood, not in this stack): have the architect
create a cron workflow dispatching a daily routine review to the
analyst.

### DIFF
--- a/crates/orchestrator/src/system_agents.rs
+++ b/crates/orchestrator/src/system_agents.rs
@@ -31,6 +31,9 @@ pub const DIAGNOSTICIAN_AGENT_NAME: &str = "agentd-diagnostician";
 /// Name of the built-in architect agent (creates agents and workflows).
 pub const ARCHITECT_AGENT_NAME: &str = "agentd-architect";
 
+/// Name of the built-in analyst agent (metrics and health analysis).
+pub const ANALYST_AGENT_NAME: &str = "agentd-analyst";
+
 /// Env var that, when set to `1`/`true`, adds remediation tools to the
 /// diagnostician's allowlist (restart_failed_agents, retry_failed_dispatches,
 /// cleanup_stale_dispatches, resolve_notification_backlog). Default off.
@@ -116,7 +119,7 @@ impl SystemAgentDef {
 /// are refreshed, and stored built-ins whose name is no longer listed here
 /// are removed as orphans.
 pub fn builtin_agent_defs() -> Vec<SystemAgentDef> {
-    vec![system_agent_def(), diagnostician_def(), architect_def()]
+    vec![system_agent_def(), diagnostician_def(), architect_def(), analyst_def()]
 }
 
 /// Definition of the `agentd-system` domain-expert agent.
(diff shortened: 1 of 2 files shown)


geoffjay (commented):
## Stack Note

PR #1273 (`sa/11-analyst`) sits at the top of a 11-PR stack on `feature/system-agents`. Two immediate concerns in the stack:

- **#1271** (`sa/09-monitor-queries`) carries a `needs-rework` label — two blocking issues flagged there (`delta()` on a counter metric and a `PromError::Unreachable` misuse for body-read failures).
- **#1272** (`sa/10-query-metrics`) is pending review.

This PR cannot merge until the whole stack resolves. The code reviewed here is self-contained and correct; this review covers this PR's diff only.

---

## Verdict: No blocking issues — recommend approval once the stack resolves

This is a clean, well-scoped addition. The implementation follows the established system-agent pattern exactly, the tool policy is correctly read-only, and the prompt is well-written with concrete baselines and playbooks.

---

## Non-blocking Notes

### 1. `query_metrics(name, ...)` — `name` reads as required in the prompt

The prompt documents the tool as `query_metrics(name, window?, range?)` and then says "Call it with no name to list the catalog." That reads as a contradiction: `name` appears positional and required but can apparently be omitted. If the `query_metrics` MCP tool (added in #1272) accepts `name` as optional, consider updating the signature in the prompt to `query_metrics(name?, window?, range?)` to avoid confusion at call time.

### 2. `approvals-resolution` gap in the catalog coverage test

The test comment says it is "covered by the approvals-backlog guidance," but the prompt's `approvals-backlog` entry only describes the backlog gauge — it does not direct the analyst to run `approvals-resolution` to confirm that resolutions are actually happening (or stalling). An analyst diagnosing sustained backlog growth would benefit from seeing both queries listed. Low priority; leaving it to the author's judgment.

### 3. `Bash(curl *)` allows curl to any URL

Appropriate for the documented fallback use case (direct Prometheus scrapes when the monitor returns 502). Worth flagging for future policy audits: this glob permits curl to any reachable endpoint, not just loopback. Scoping to `Bash(curl http://127.0.0.1:*/*)` would be more surgical when the project tightens Bash allowances.

---

## What Works Well

- **Tool policy** is correctly scoped: all `get_*` / `list_*` tools in the MCP catalog are genuinely read-only; the two escalation channels (`create_notification`, `send_room_message`) are explicit exact-matches rather than wildcards; `send_agent_message` and all write/lifecycle tools are correctly absent.
- **`analyst_prompt_covers_the_query_catalog` test** is excellent hygiene — it will catch future catalog additions that are not reflected in the prompt, cross-crate without a compile-time import.
- **`analyst_policy_is_read_only_plus_escalation`** covers a comprehensive allow/deny matrix; the bare `Bash` (no args) denial correctly verifies that Bash without the `curl` qualifier is blocked.
- **Prompt structure** — baselines with deviation guidance and four named playbooks (cost spike, latency, failures, routine review) give the agent clear decision trees without over-prescribing.
- Integration test count bump (3 to 4) is correct and the existing idempotency assertions still hold.

PR #1274, opened 2026-06-12
feat(ui): monitor named-query client and live System Resources charts

geoffjay:
Wire the monitoring page to the monitor service, replacing the
"Coming Soon" placeholder section that shipped before the service
existed.

- MonitorClient gains listQueries() and runQuery(name, {window, mode,
  rangeMinutes, stepSecs}) for the curated PromQL catalog, with types
  (NamedQuery, PromData vector/matrix/scalar, QueryResult) mirroring
  the Rust API. A 502 (Prometheus down) surfaces as ApiError(502),
  distinct from the monitor itself being unreachable.
- New ResourceTrendCard (compact time-series card) and DiskUsageCard
  (per-mount fill bars with capacity colour shifts) components.
- MonitoringDashboard's System Resources section now renders live CPU,
  memory, disk, and load-average data from the monitor history ring
  buffer via useSystemMetrics, with quiet unavailable/empty states and
  a header badge when the monitor is down. The stale "port 17003"
  hardcoded hint is gone.
- PlaceholderChart is no longer used by any page (file retained;
  removal blocked by the destructive-command hook).
- MSW handlers/factories for /queries and /queries/{name}.

### DIFF
--- a/ui/src/components/monitoring/DiskUsageCard.tsx
+++ b/ui/src/components/monitoring/DiskUsageCard.tsx
@@ -1,0 +1,130 @@
+/**
+ * DiskUsageCard — per-mount disk utilisation bars from the latest monitor
+ * snapshot.
+ *
+ * Time series per mount point is noisy and rarely actionable; the current
+ * fill level per mount is what operators check. Bars colour-shift as they
+ * approach capacity.
+ */
+
+import { HardDrive } from "lucide-react";
+import { ChartSkeleton } from "@/components/common/LoadingSkeleton";
+import type { DiskMetrics } from "@/types/monitor";
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+export interface DiskUsageCardProps {
+	disks: DiskMetrics[];
+	/** False when the monitor service is unreachable */
+	available?: boolean;
+	loading?: boolean;
+	/** Height of the content area in pixels (default 140) */
+	height?: number;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function formatBytes(bytes: number): string {
+	const gib = bytes / 1024 ** 3;
+	if (gib >= 1024) return `${(gib / 1024).toFixed(1)} TiB`;
+	return `${gib.toFixed(0)} GiB`;
+}
+
+/** Bar colour by fill level: calm → warning → critical. */
+function barColor(usagePercent: number): string {
+	if (usagePercent >= 90) return "var(--th-status-error-text, #ef4444)";
+	if (usagePercent >= 75) return "var(--th-status-warning-text, #f59e0b)";
+	return "#3b82f6";
+}
+
+// ---------------------------------------------------------------------------
+// Component
+// ---------------------------------------------------------------------------
+
+export function DiskUsageCard({
+	disks,
+	available = true,
+	loading = false,
+	height = 140,
+}: DiskUsageCardProps) {
+	const hasData = available && disks.length > 0;
+
+	return (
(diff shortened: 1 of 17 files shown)


geoffjay (commented):
## Review: Monitor Named-Query Client + Live System Resources Charts

**Stack position**: `sa/12-ui-monitor-queries` branches directly from `main`. PR #1275 (`sa/13-ui-platform-metrics`) is stacked on top of this — merging #1274 is the prerequisite for landing #1275.

---

### What's good

- **`PromData` discriminated union** (`types/monitor.ts`) faithfully mirrors the Prometheus wire format. The `resultType` tag makes exhaustive switching safe and the `scalar` variant is included even though it is not consumed yet.
- **`encodeURIComponent` on the query name** in `runQuery` — verified by the dedicated URL-encoding test. The right thing to do given the catalog uses hyphenated names.
- **Param serialization** (`camelCase` to `snake_case`) is correct and fully tested: zero params on a bare call, all four params on a populated `RunQueryOptions`, 502 surfaces as `ApiError(502)`, 404 carries the catalog hint.
- **`DiskUsageCard` accessibility** — `progressbar` role with `aria-valuenow / aria-valuemin / aria-valuemax` per disk bar is well done and tested.
- **Three-state handling** (loading vs unavailable vs empty-but-available vs data) in both new components is robust and never crashes the page.
- **`beforeEach` + `mockUseSystemMetrics.mockReturnValue` refactor** in the dashboard test is a clear improvement over a single static mock. The refresh-interval test fix (scoping to `getByRole("button")`) removes the false ambiguity with Load Average readout labels.
- **`PlaceholderChart` removal from the barrel export** is clean; retaining the file itself is the right call given the destructive-command hook constraint.

---

### Non-blocking suggestions

**1. Missing "15m" readout in Load Average (`MonitoringDashboard.tsx`)**

The 15m series is added to the chart `series` array but there is no matching entry in `readouts`. The yellow line renders in the chart but has no legend dot or current-value label. The user cannot identify that line without a corresponding readout. If intentional (space saving), a comment would help; otherwise the 15m readout should be added alongside "1m" and "5m".

**2. `useSystemMetrics()` polling rate is decoupled from the dashboard refresh selector**

The hook takes no interval argument (confirmed from its signature), so system resource cards refresh at a fixed internal rate regardless of the user's selected refresh cadence. `useUsageMetrics` and `useMetrics` both accept the user-selected `refreshInterval`; `useSystemMetrics` does not. This is a constraint of the hook's design, not a defect in this PR, but it is a noticeable inconsistency: selecting "10s" does not make the CPU / memory charts update at 10s. A follow-up to add an optional interval parameter to the hook, and a JSDoc note in the meantime, would help.

**3. Recurring hardcoded `#3b82f6` (same pattern in child PR #1275)**

`DiskUsageCard.barColor` returns the raw hex `"#3b82f6"` for the normal (below 75%) bar fill while the warning and error levels use CSS variables with fallbacks. Both PRs in this stack share the same asymmetry — it is a codebase-wide pattern rather than a one-off. A follow-up using `var(--th-accent, #3b82f6)` across the monitoring components would make the palette fully theme-driven.

**4. `formatBytes` floor-clamps sub-GiB values to "0 GiB"**

Values below 1 GiB are displayed as "0 GiB". Unlikely to matter for physical disk mounts, but tmpfs or RAM-backed mounts can be smaller. Not blocking for the current use case — if it comes up, add a MiB branch before the GiB return.

---

### Stack impact

Landing this PR directly unblocks #1275. The `monitorClient.runQuery` method and the `QueryResult` / `VectorSample` types added here are the foundation that `usePlatformMetrics` in #1275 imports. #1275 has already been reviewed and is waiting on this parent.

PR #1275, opened 2026-06-12
feat(ui): Platform Metrics section from the monitor named-query catalog

geoffjay:
Add a Prometheus-backed Platform Metrics section to the monitoring page,
between the system health panel and the token-usage charts.

- usePlatformMetrics polls nine catalog queries (dispatch success rate
  and throughput, agent restarts, approvals backlog, HTTP error rate
  and p95 latency per service, session cost, agents-active vs
  websocket-connections) and distinguishes monitorDown (service
  unreachable) from prometheusDown (502) so the page can say which
  layer broke. Exported vectorEntries/vectorValue helpers parse
  instant-vector results.
- PlatformMetricsSection renders tone-coded stat cards (dispatch
  success warns below 90%, an agents/connections mismatch hints at
  diagnose_state_mismatches, nonzero backlogs and restarts warn) and
  per-service bar lists with p95 converted to ms and NaN rows (no
  traffic) filtered out.
- Degraded banner: "Prometheus is offline — platform metrics
  unavailable (system resources are still live)" vs monitor
  unreachable.

### DIFF
--- a/ui/src/components/monitoring/PlatformMetricsSection.tsx
+++ b/ui/src/components/monitoring/PlatformMetricsSection.tsx
@@ -1,0 +1,287 @@
+/**
+ * PlatformMetricsSection — Prometheus-backed platform health from the
+ * monitor service's curated named-query API.
+ *
+ * Stat cards (dispatch success, agents vs connections, approvals backlog,
+ * session cost, restarts) over per-service HTTP error-rate and p95-latency
+ * bar lists. Renders explicit degraded states that distinguish "Prometheus
+ * offline" (system resources still live) from "monitor unreachable".
+ */
+
+import { AlertTriangle } from "lucide-react";
+import { ChartSkeleton } from "@/components/common/LoadingSkeleton";
+import type {
+	PlatformQueryResults,
+	UsePlatformMetricsResult,
+} from "@/hooks/usePlatformMetrics";
+import { vectorEntries, vectorValue } from "@/hooks/usePlatformMetrics";
+
+// ---------------------------------------------------------------------------
+// Stat card
+// ---------------------------------------------------------------------------
+
+export type StatTone = "neutral" | "ok" | "warn" | "error";
+
+export interface PlatformStatCardProps {
+	label: string;
+	/** Pre-formatted display value; "—" when no data */
+	value: string;
+	/** Small line under the value (e.g. the window or a breakdown) */
+	hint?: string;
+	tone?: StatTone;
+}
+
+const TONE_CLASSES: Record<StatTone, string> = {
+	neutral: "text-th-text",
+	ok: "text-th-status-success-text",
+	warn: "text-th-status-warning-text",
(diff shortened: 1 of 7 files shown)


geoffjay (commented):
## Review: Platform Metrics Section

**Stack note**: This PR is stacked on #1274 (`sa/12-ui-monitor-queries`), which is still open and carries the `review-agent` label. PR #1275 depends on `monitorClient.runQuery` and the `QueryResult`/`VectorSample` types introduced by that parent. The code here is correct; the `merge-queue` label is added with the understanding the conductor should not land this until #1274 merges first.

---

### What's good

- The `monitorDown` / `prometheusDown` split is well-reasoned. The invariant holds: `sawBadGateway = true` proves the monitor responded, so only `\!sawSuccess && \!sawBadGateway` correctly means "nothing answered at all." The test that uses 404s (not 502s) to sidestep the retry-backoff penalty is a nice touch.
- `Promise.allSettled` over all nine queries means a single flapping query does not drop the rest of the dashboard.
- NaN filtering for latency (`Number.isFinite`) is correct: `histogram_quantile` returns NaN for services with no traffic, and rendering those rows as `0 ms` would be incorrect.
- `hasLoadedRef` ensures the skeleton only shows on first load, not on every background poll.
- `MonitoringDashboard.test.tsx` mock correctly preserves the real vector helpers (`...original`) so unit tests for components that import them still get the real implementations.
- Test coverage is thorough: happy path, both degraded states, mismatch hint, NaN filtering, and the `vectorValue` sum behaviour are all exercised.

---

### Non-blocking suggestions

**1. `fetch` callback name shadows the global (`usePlatformMetrics.ts`)**

```ts
// current
const fetch = useCallback(async () => { ... }, []);
```

`fetch` is a global in every browser/Node context. Naming the inner callback the same makes intent slightly ambiguous and could confuse linters. Consider `fetchMetrics` or `loadMetrics`.

**2. Latency `fraction` prop has redundant `/1` and a saturation cliff (`PlatformMetricsSection.tsx`)**

```tsx
fraction={(v) => v / 1}   // equivalent to (v) => v
```

`v / 1` is identity; the division looks like an incomplete normalisation (`/ maxValue`). Suggest removing it.

Separately, with this fraction and `Math.min(100, ... * 100)`, any service whose p95 exceeds 1 s shows a full-width bar, making two services at 1.2 s and 3 s visually identical. Not blocking for the initial implementation, but worth a follow-up that clamps against a configurable max (e.g. 2 s).

**3. Hardcoded bar fill colour (`PlatformMetricsSection.tsx`)**

```tsx
background: warn
  ? "var(--th-status-warning-text, #f59e0b)"
  : "#3b82f6",           // does not follow theming
```

The warn path uses a CSS variable with fallback; the normal path uses a raw hex. For theme-consistency, consider `var(--th-accent, #3b82f6)` for parity.

---

Overall this is a clean addition with solid error-surface coverage and good tests. Approving.

PR #1287, opened 2026-06-14
feat(cli): route all commands through core gateway with bearer tokens

### LINKED ISSUE #1117: CLI auth integration: route all commands through core gateway with bearer tokens
## Context

The CLI's `agent auth login` stores a session token at `~/.config/agentd/session`, but all other commands (orchestrator, notify, ask, wrap, memory, communicate, index, project, apply, status) talk directly to individual services with no auth headers. This issue makes every CLI command load the stored session and route through the core gateway.

## Checklist

- [ ] **Add bearer token support to all Rust service clients** -- Each service client (OrchestratorClient, NotifyClient, AskClient, WrapClient, MemoryClient, CommunicateClient, IndexClient) gets a `token: Option<String>` field and injects `.bearer_auth(token)` on every HTTP request when present. Existing `from_env()` constructors are preserved so the MCP server can use them without auth.
  - Files: `crates/orchestrator/src/client.rs`, `crates/notify/src/client.rs`, `crates/ask/src/client.rs`, `crates/memory/src/client.rs`, `crates/communicate/src/client.rs`, `crates/cli/src/commands/index.rs` (inline IndexClient)

- [ ] **Route all CLI commands through core gateway with auth** -- Refactor `crates/cli/src/main.rs` command dispatch so every command reads `AGENTD_CORE_SERVICE_URL` (default `http://localhost:17000`), loads the session token, and constructs the service client pointing at `{core_url}/api/v1/{service}` with the token. Add helpers: `gateway_url(service)` to build the gateway path, and `load_token_or_warn()` that loads the token and prints a warning if absent but doesn't hard-fail. The `Status` command should use the core gateway's aggregate health endpoint at `/api/v1/health`.
  - Files: `crates/cli/src/main.rs`

## Dependencies

- Foundation issue (gateway must proxy all services before CLI can route through it)

geoffjay:
Routes all CLI service commands through the core gateway at `http://localhost:17000/api/v1/{service}` and injects a bearer token loaded from `~/.config/agentd/session`. Each service client gains a `with_token()` constructor; the CLI loads the session token at startup and warns if absent.

Closes #1117


geoffjay (commented):
## Review: feat(cli): route all commands through core gateway with bearer tokens

### Stack context

This PR is stacked directly on #1129 (feat(auth): gateway completeness) which is still open. **Cannot merge until #1129 lands.** Three child PRs (#1288, #1289, #1290) build on this one — they each have `needs-rework` from separate reviews, so there is no merge-blocking pressure from above yet.

---

The implementation is mechanically correct. The `with_token()` builder is applied uniformly across all six client crates (ask, communicate, memory-api, notify, orchestrator, wrap). Every HTTP verb — GET, POST, PUT, PATCH, DELETE — gets the conditional header injection. The `new_no_proxy` constructor in `CommunicateClient` correctly gains `token: None`. The `Prompt` command correctly clones the token when sharing it between two clients. `load_token()` already writes files with 0600 permissions, so the token-at-rest story is handled.

---

### Non-blocking suggestions

**1. Silent removal of per-service env vars**

The old code honored `AGENTD_NOTIFY_SERVICE_URL`, `AGENTD_ASK_SERVICE_URL`, `AGENTD_ORCHESTRATOR_SERVICE_URL`, etc. These are now silently ignored — only `AGENTD_CORE_SERVICE_URL` matters. Developers who run individual services without the gateway (common in service-level testing or partial-stack dev setups) now have no bypass.

Consider a fallback check in `gateway_url`:

```rust
fn gateway_url(service: &str) -> String {
    // Allow per-service override for direct-access dev/test environments
    let env_key = format!("AGENTD_{}_SERVICE_URL", service.to_uppercase());
    if let Ok(direct) = env::var(&env_key) {
        return direct;
    }
    let core = env::var("AGENTD_CORE_SERVICE_URL")
        .unwrap_or_else(|_| "http://localhost:17000".to_string());
    format!("{}/api/v1/{}", core, service)
}
```

If gateway-first is a hard requirement and direct access is intentionally removed, that decision is worth a sentence in `CLAUDE.md` or a warning when those old env vars are detected but ignored.

---

**2. Repeated boilerplate in `main.rs`**

The 10+ command arms all follow the same pattern:

```rust
let token = load_token_or_warn();
let url = gateway_url("service");
let mut client = ServiceClient::new(url);
if let Some(t) = token {
    client = client.with_token(t);
}
```

This is harmless now but will grow stale if the pattern ever changes. A small macro or a generic helper would eliminate the duplication and make future changes (e.g. adding a timeout, request ID, etc.) a one-line edit rather than ten.

---

**3. No tests for token injection**

The `with_token` method and the conditional `bearer_auth` injection path have zero test coverage. A unit test using `mockito` or `wiremock` verifying that:
- A client built with `.with_token("abc")` sends `Authorization: Bearer abc` on every verb
- A client built without a token sends no `Authorization` header

would catch any future regression. This is especially important given the security implications of accidentally omitting the header.

---

**4. `CommunicateClient::new_no_proxy` has no token path**

`new_no_proxy` is updated to initialize `token: None`, but there is no `with_token()` / `new_no_proxy_with_token()` to attach a token afterwards. In practice the CLI uses `new()`, not `new_no_proxy`, so this is a latent gap rather than an active bug. Worth a `// TODO: expose with_token() for no-proxy callers` comment so it gets picked up if someone adds gateway routing for the no-proxy path.

---

### What's well done

- The `with_token(mut self, ...)` builder pattern is consistent and idiomatic across all clients.
- `gateway_url` is a clean, single-responsibility function — easy to understand and modify.
- `load_token_or_warn` gracefully degrades: commands still work without a token (the gateway 401 surfaces naturally), and the user gets actionable guidance.
- Removing the old per-service env var boilerplate from `main.rs` significantly reduces noise.
- The `Prompt` command correctly clones the token without double-loading from disk.

PR #1288, opened 2026-06-14
feat(ui): login page, auth store, gateway routing, WebSocket auth

### LINKED ISSUE #1118: UI auth integration: login page, auth store, gateway routing, WebSocket auth
## Context

The React UI has no login page, no auth context, and no token management. The `ApiClient` base class in `ui/src/services/base.ts` already has an `onRequest` hook (line 28-29) designed for future auth header injection. This issue adds the full auth flow to the UI: login/register, token storage in localStorage, auth-gated routing, and auth headers on all API calls.

## Checklist

- [ ] **Create auth store and core API client** -- Add a Zustand auth store (matching existing settingsStore/themeStore/toastStore pattern) with: `token` loaded from localStorage key `agentd_token`, `user` and `activeOrganization` state, and actions for `login()`, `register()`, `logout()`, `checkSession()`, `switchOrganization()`. Add a core auth API client extending `ApiClient` for the auth endpoints.
  - Files: `ui/src/stores/authStore.ts` (new), `ui/src/services/auth.ts` (new), `ui/src/services/config.ts`

- [ ] **Add login page and auth-gated routing** -- Create `LoginPage` and `RegisterPage` components. Create `RequireAuth` wrapper that checks `authStore.isAuthenticated`, redirects to `/login` if not, calls `checkSession()` on mount. Restructure `App.tsx` routes so login/register are outside the shell and everything else is inside `RequireAuth`. Add user display and logout button to the header.
  - Files: `ui/src/pages/LoginPage.tsx` (new), `ui/src/pages/RegisterPage.tsx` (new), `ui/src/components/common/RequireAuth.tsx` (new), `ui/src/App.tsx`, `ui/src/layouts/Header.tsx`

- [ ] **Route all UI API calls through core gateway with auth headers** -- Change `ui/src/services/config.ts` so all service URLs route through the core gateway (e.g., `orchestratorServiceUrl = ${coreServiceUrl}/api/v1/orchestrator`). Use the existing `onRequest` hook on `ApiClient` to inject `Authorization: Bearer <token>` from localStorage. Create a shared `withAuth` helper to avoid repeating the injection logic across all client singletons.
  - Files: `ui/src/services/config.ts`, all service client files in `ui/src/services/`

- [ ] **Handle 401 responses globally** -- When any API call returns 401, clear the auth store and redirect to `/login`. Implement via an `onUnauthorized` callback in `ApiClient` options or by checking response status in `parseResponse()`.
  - Files: `ui/src/services/base.ts`, `ui/src/stores/authStore.ts`

- [ ] **WebSocket auth for UI connections** -- Orchestrator agent streaming and communicate chat use WebSocket connections that bypass the gateway. Modify `openWebSocket()` in `base.ts` and WS hooks to append `?token=<bearer>` to the WS URL.
  - Files: `ui/src/services/base.ts`, `ui/src/hooks/useAgentStream.ts`, `ui/src/hooks/useCommunicateSocket.ts`, `ui/src/hooks/useWebSocket.ts`

## Dependencies

- Foundation issue (gateway completeness, auth store depends on core service URL)
- Service tenant scoping issue (server-side WS auth needed before UI WS auth works)

geoffjay:
Adds a Zustand auth store with localStorage persistence, login/register pages, a RequireAuth route guard, and routes all UI service calls through the core gateway with bearer token injection. WebSocket connections include the session token as a query parameter.

Closes #1118


geoffjay (commented):
## Review: feat(ui): login page, auth store, gateway routing, WebSocket auth

### Stack context

This PR is stacked on #1287 (feat(cli): bearer-token gateway routing) → #1129 (feat(auth): gateway completeness) — both still open. **This PR cannot merge until those land.** The stacking itself is clean; the diff applies correctly on top of its parent. No restack needed.

---

### Blocking issues

**1. `coreServiceUrl` is duplicated between `auth.ts` and `config.ts`**

`services/auth.ts` defines its own resolution chain for `coreServiceUrl`:

```ts
// auth.ts
const coreServiceUrl =
    runtimeServiceUrl("core") ??
    import.meta.env.VITE_AGENTD_CORE_SERVICE_URL ??
    "http://localhost:17000";
```

`services/config.ts` now exports exactly the same value under `serviceConfig.coreServiceUrl`. `auth.ts` should consume that instead:

```ts
import { serviceConfig } from "./config";
// use serviceConfig.coreServiceUrl everywhere in this file
```

If the resolution order or default port ever changes, it will need updating in two places and they will silently diverge.

---

**2. `TOKEN_KEY` / `"agentd_token"` magic string is not shared**

`authStore.ts` defines `const TOKEN_KEY = "agentd_token"` (unexported), but `auth.ts` references `"agentd_token"` as a raw string literal:

```ts
// auth.ts - logout()
const token = localStorage.getItem("agentd_token");
```

Export `TOKEN_KEY` from `authStore.ts` and import it in `auth.ts`. A typo or future rename will otherwise diverge silently at runtime.

---

### Non-blocking suggestions

**3. 401 handler does not early-exit after scheduling redirect** (`services/base.ts`)

```ts
if (response.status === 401) {
    localStorage.removeItem("agentd_token");
    window.location.href = "/login";   // schedules navigation; does NOT stop execution
}

let message = `HTTP ${response.status}`;
// ... throw ApiError(401, ...)
```

`window.location.href` is asynchronous — JS keeps running. The code falls through, constructs an error, and throws it, which callers catch and render as an error toast/banner before the redirect completes. Suggested fix:

```ts
if (response.status === 401) {
    localStorage.removeItem("agentd_token");
    window.location.href = "/login";
    throw new ApiError(401, "Session expired");
}
```

---

**4. `RequireAuth.tsx` discards the original URL** (line 16)

```tsx
return <Navigate to="/login" replace />;
```

Users navigating directly to `/agents/123` get sent to `/` after login, not back to the page they wanted. The React Router convention is:

```tsx
const location = useLocation();
return <Navigate to="/login" state={{ from: location }} replace />;
```

And in `LoginPage.tsx`/`RegisterPage.tsx`:

```ts
const location = useLocation();
const from = (location.state as { from?: Location })?.from?.pathname ?? "/";
navigate(from, { replace: true });
```

Not blocking by itself, but it's a UX regression from the current un-guarded state and will generate follow-up tickets.

---

**5. WebSocket token should be URL-encoded** (`services/base.ts` ~line 286)

```ts
const wsUrl = token
    ? `${url}${url.includes("?") ? "&" : "?"}token=${token}`
    : url;
```

Standard JWTs are Base64url-safe, but `encodeURIComponent(token)` is defensively correct and makes the code resilient to different token formats that might arrive from the server.

---

**6. Login/Register pages use raw inline styles**

`LoginPage.tsx` and `RegisterPage.tsx` rely entirely on inline `style={{...}}` objects. If the rest of the UI uses a design system (Tailwind, CSS Modules, etc.) these pages will visually diverge. Suggest following whatever styling pattern the existing pages use, even as a quick pass. If inline styles are intentional for an MVP bootstrap, a brief comment would help.

---

### What's well done

- `withAuth` hook design is clean — a single injection point in `ApiClientOptions.onRequest`, no monkey-patching.
- All six service clients are updated uniformly; nothing is missed.
- `RequireAuth` correctly wraps `AppShell` rather than individual routes, keeping the nesting clean.
- The `AuthApiClient` is correctly excluded from `withAuth` (avoids circular auth injection on login).
- Exporting `getStoredToken()` from the store gives service code a stable accessor without importing the hook outside React.

Please address items 1 and 2 (DRY violations / shared constant) before merge. The other items are noted for follow-up.

PR #1289, opened 2026-06-14
feat: add organization_id tenant scoping to downstream services

### LINKED ISSUE #1119: Downstream service tenant scoping: add organization_id and WebSocket auth
## Context

No service currently scopes data by organization. All data is globally accessible. This issue adds `organization_id` columns to service databases, scopes all queries using the `OptionalTenantId` extractor from `agentd-common`, and adds server-side WebSocket auth for orchestrator and communicate.

Using `OptionalTenantId` (not required `TenantId`) so services still work without the gateway in dev/test. Hook and monitor stay system-wide -- they serve infrastructure purposes and don't hold per-org data.

## Checklist

- [ ] **Orchestrator: add organization_id and scope queries (reference implementation)** -- This sets the pattern for all other services. New migration adds `organization_id TEXT` to agents, workflows, dispatch_log, and projects tables. Update entity structs to include `organization_id: Option<String>`. Update storage queries: `list` methods filter by org_id when present, `create` methods set org_id on insert. Update API handlers to extract `OptionalTenantId` from request and pass to storage.
  - Files: `crates/orchestrator/src/migration/` (new), `crates/orchestrator/src/entity/*.rs`, `crates/orchestrator/src/storage.rs`, `crates/orchestrator/src/api.rs`

- [ ] **Notify: add organization_id and scope queries** -- Same pattern as orchestrator. Add `organization_id` to notifications table.
  - Files: `crates/notify/src/migration/`, `crates/notify/src/entity/`, `crates/notify/src/storage.rs`, `crates/notify/src/api.rs`

- [ ] **Communicate: add organization_id and scope queries** -- Rooms get `organization_id`. Messages and participants inherit scope from their room.
  - Files: `crates/communicate/src/migration/`, `crates/communicate/src/entity/`, `crates/communicate/src/storage.rs`, `crates/communicate/src/api.rs`

- [ ] **Memory, ask, index: add organization_id and scope queries** -- Same pattern for remaining data-holding services.
  - Files: `crates/memory/`, `crates/ask/`, `crates/index/` -- migration, entity, storage, api modules in each

- [ ] **WebSocket auth for orchestrator and communicate (server-side)** -- Add shared WS token validation helper to `agentd-common` (`validate_ws_token` calls `GET {core_url}/auth/me` with bearer token, returns `(user_id, org_id)`). Apply to both WS handlers: extract `token` from query params, validate before upgrade, reject with 401 if invalid.
  - Files: `crates/common/src/ws_auth.rs` (new), `crates/orchestrator/src/websocket.rs`, `crates/communicate/src/websocket.rs`

## Dependencies

- Foundation issue (X-Tenant-ID extractor, token validation endpoint)

geoffjay:
Adds SeaORM migrations to add `organization_id TEXT` to all tenant-scoped tables across orchestrator, notify, communicate, memory-api, and ask services. API handlers extract `OptionalTenantId` from the gateway-injected header and filter queries accordingly (matching org OR NULL for legacy rows). Adds a WS token validation helper in agentd-common.

Closes #1119


geoffjay (commented):
## Review: feat: add organization_id tenant scoping to downstream services

### Stack context

This PR is stacked on #1288 (UI auth) -> #1287 (CLI gateway routing) -> #1129 (auth gateway completeness), all open. **Cannot merge until parents land.** Child PR #1290 (backfill-tenant command) depends on this one.

The overall scaffolding is consistent: entity field, idempotent migration, paired `_org` storage method, handler wired to `OptionalTenantId`. The test updates to `WorkflowConfig` constructors are all correct.

---

### Blocking issues

**1. All list filters use strict equality — contradicts the PR description and `tenant.rs` policy**

Every `_org` storage method across all services filters with:

```rust
if let Some(oid) = org_id {
    query = query.filter(Column::OrganizationId.eq(oid));
}
```

The PR description states queries should match "org OR NULL for legacy rows." The `tenant.rs` module (updated in the child PR) documents this as the required pattern during the transition period:

```sql
WHERE organization_id = ? OR organization_id IS NULL
```

With strict equality, all pre-migration rows (where `organization_id IS NULL`) are invisible to every authenticated tenant the moment this PR deploys — until `agent admin backfill-tenant` is run manually. The backfill is a separate, operator-driven step; the system should degrade gracefully without it.

Fix for each filter site (SeaORM `Condition::any` approach):

```rust
if let Some(oid) = org_id {
    query = query.filter(
        Condition::any()
            .add(Column::OrganizationId.eq(oid))
            .add(Column::OrganizationId.is_null()),
    );
}
```

This applies to all six list/paginated methods:
- `ask/storage.rs` — `list_org`
- `communicate/storage.rs` — `list_rooms_org`
- `notify/storage.rs` — `list_paginated_org`
- `orchestrator/storage.rs` — `list_org`, `list_paginated_org`
- `orchestrator/scheduler/storage.rs` — `list_workflows_org`, `list_workflows_paginated_org`
- `orchestrator/storage.rs` (`ProjectStorage`) — `list_org`

---

**2. `ask` service has a broken create/list pair**

`create_question_handler` uses `OptionalTenantId(_org_id)` (underscore — intentionally unused) and the storage always writes:

```rust
organization_id: Set(None),
```

But `list_questions_handler` uses `org_id.as_deref()` with strict equality filtering. Any question created via the gateway has `org_id = NULL`, so listing as an authenticated tenant returns zero results — the authenticated user creates questions they can never list.

Either:
- Store the org_id on creation (remove the underscore, pass it to the insert), or
- Also apply the OR NULL pattern so NULL-created questions are visible to all tenants

The first option is more correct; it prevents cross-tenant leakage. If the `_org_id` is intentional (phase 1: column only, filtering later), the `list_org` filter must also be reverted to pass `None` until creates are wired up.

---

**3. `create_agent` org assignment is non-atomic** (`orchestrator/src/api.rs` line ~278)

```rust
let agent = state.manager.spawn_agent(req.name, config, false).await?;

if org_id.is_some() {
    state.manager.agent_storage()
        .set_agent_organization(&agent.id, org_id.as_deref())
        .await
        .map_err(ApiError::Internal)?;
}
```

`spawn_agent` writes the agent to the DB with `organization_id = NULL`, then a second UPDATE sets the org. Between these two calls the agent is observable without its org scope (a concurrent `list_agents_paginated_org` call would miss it). The `Agent` struct already has an `organization_id` field initialized to `None` in `Agent::new()`. The proper fix is:

```rust
// Before calling spawn_agent:
let mut req = req;
req.organization_id = org_id;  // or however spawn_agent surfaces this

// If spawn_agent doesn't forward org_id, set it on the agent after creation
// but before the DB insert — not as a separate UPDATE.
```

Alternatively, pass `org_id` directly to the `AgentStorage::insert` path so the initial INSERT includes the correct value, avoiding the two-step entirely.

---

### Non-blocking suggestions

**4. `ws_auth::validate_ws_token` creates a new `reqwest::Client` per call** (`common/src/ws_auth.rs` line 205)

```rust
let client = reqwest::Client::new();
```

`reqwest::Client` initializes a full connection pool on construction. Under concurrent WebSocket upgrades this allocates a new pool per handshake. Use a shared static client:

```rust
use once_cell::sync::Lazy;
static HTTP_CLIENT: Lazy<reqwest::Client> = Lazy::new(reqwest::Client::new);
```

---

**5. Silent `"unknown"` user_id on malformed auth response** (`common/src/ws_auth.rs` line 227)

```rust
user_id: body["id"].as_str().unwrap_or("unknown").to_string(),
```

If the `/auth/me` response is missing the `"id"` field the function succeeds but returns `"unknown"`. If any downstream code uses `WsAuthInfo.user_id` for audit logging or access control, this is a silent data quality bug. Return an error instead:

```rust
user_id: body["id"]
    .as_str()
    .ok_or_else(|| anyhow::anyhow!("auth/me response missing 'id' field"))?
    .to_string(),
```

---

**6. Memory service tenant scoping incomplete**

`memory/src/api.rs` registers the extractor with `_org_id` on both `create_memory` and `list_memories`, and `memory/src/storage.rs` hardcodes `organization_id: Set(None)`. The column is added via migration but neither filtering on read nor scoping on write is implemented.

If this is intentional (phase 1: add column, wire up filtering in a follow-up), a `// TODO(#NNNN): tenant-scope memory queries` comment would prevent it from looking like an oversight in future reviews.

---

### What's well done

- Idempotent migrations using `"duplicate column name"` guard are consistent and safe across all five services.
- `OptionalTenantId` is correctly used (not `TenantId`) so services continue to work without the gateway in dev/test.
- `down()` migrations explicitly document the SQLite < 3.35 `DROP COLUMN` limitation.
- The `_org` paired-method pattern (original method delegates to `_org(None)`) keeps backwards compatibility for callers that don't need tenant filtering.
- Test constructors updated correctly for the new `organization_id: None` field.

Please address items 1-3 before merge. Items 4-6 are noted for follow-up.

PR #1290, opened 2026-06-14
feat(cli): add admin backfill-tenant command for organization_id migration

### LINKED ISSUE #1120: Data migration: backfill existing unscoped data with organization_id
## Context

After the service tenant scoping work adds nullable `organization_id` columns, existing data will have `NULL` values. This issue provides a strategy to backfill that data and handle the transition period gracefully.

## Checklist

- [ ] **Backfill strategy and CLI command** -- If only one organization exists in the system, automatically assign all NULL rows to it. Otherwise, provide `agent admin backfill-tenant --org-id <id>` CLI command that updates all NULL `organization_id` values across service databases. The `OptionalTenantId` query layer includes NULL rows during the transition period so existing data remains accessible before backfill.
  - Files: `crates/cli/src/commands/admin.rs` (new), `crates/cli/src/main.rs`

- [ ] **Document NULL handling policy** -- During the transition, queries with a tenant ID should include rows where `organization_id` matches OR is NULL (legacy data). This policy should be consistent across all services and documented in the common tenant module.
  - Files: `crates/common/src/tenant.rs` (doc comments)

## Dependencies

- Service tenant scoping issue (schemas must have the columns before backfill can run)

geoffjay:
Adds `agent admin backfill-tenant --org-id <id>` to assign `organization_id` to all NULL rows across service databases (orchestrator, notify, communicate, memory) after the tenant scoping migration. Supports `--dry-run` for previewing affected row counts. Documents the NULL handling policy in the common tenant module.

Closes #1120

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,7 @@ dependencies = [
  "hyper",
  "metrics",
  "metrics-exporter-prometheus",
+ "notify",
  "rand 0.8.5",
  "reqwest",
  "sea-orm",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1118,6 +1118,7 @@ dependencies = [
  "orchestrator",
  "prettytable-rs",
  "reqwest",
+ "sea-orm",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -26,6 +26,7 @@ crossterm = { version = "0.28", features = ["event-stream"] }
 bytes = "1"
 serde_yaml = "0.9"
 toml = { workspace = true }
+sea-orm = { version = "1.1", features = ["sqlx-sqlite", "runtime-tokio-rustls", "macros"] }
 urlencoding = "2.1"
 dialoguer = "0.11"
 dirs = "5"
--- a/crates/cli/src/commands/admin.rs
+++ b/crates/cli/src/commands/admin.rs
@@ -24,7 +24,7 @@ use agentd_install::migrate::DB_SERVICES;
 use anyhow::{Context, Result};
 use clap::Subcommand;
 use colored::*;
-use sea_orm::ConnectionTrait;
+use sea_orm::{ConnectionTrait, Statement, Value};
 
 /// Admin subcommands.
 #[derive(Debug, Subcommand)]
@@ -93,6 +93,9 @@ const SCOPED_TABLES: &[ServiceTables] = &[
 // ---------------------------------------------------------------------------
 
 /// Count rows with NULL organization_id in a table.
+///
+/// `table` must come from the hardcoded `SCOPED_TABLES` constant — it is
+/// never user-supplied and is safe to interpolate into the statement string.
 async fn count_null_org_rows(
     db: &sea_orm::DatabaseConnection,
     table: &str,
@@ -104,8 +107,8 @@ async fn count_null_org_rows(
         cnt: i64,
     }
 
-    let result = CountRow::find_by_statement(sea_orm::Statement::from_string(
-        sea_orm::DatabaseBackend::Sqlite,
+    let result = CountRow::find_by_statement(Statement::from_string(
+        db.get_database_backend(),
         format!("SELECT COUNT(*) AS cnt FROM {} WHERE organization_id IS NULL", table),
     ))
     .one(db)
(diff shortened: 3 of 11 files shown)


geoffjay (commented):
## Review: feat(cli): add admin backfill-tenant command for organization_id migration

### Stack context

This PR sits on top of a four-deep stack: #1289 (org-id tenant scoping) -> #1288 (UI auth) -> #1287 (CLI gateway routing) -> #1129 (gateway completeness). All four parents are open. **This PR cannot merge until the full stack below it lands.** No restack needed -- the branch applies cleanly on its parent.

\`SCOPED_TABLES\` correctly covers all four services in \`DB_SERVICES\` (memory, notify, orchestrator, communicate). The overall structure -- dry-run count then live update, per-service loop, JSON/human output modes -- is solid.

---

### Blocking issue

**SQL injection in \`update_null_org_rows\`** (\`crates/cli/src/commands/admin.rs\` ~line 116)

\`\`\`rust
let stmt = format!(
    "UPDATE {} SET organization_id = '{}' WHERE organization_id IS NULL",
    table, org_id
);
let result = db.execute_unprepared(&stmt).await?;
\`\`\`

\`org_id\` is user-supplied CLI input interpolated verbatim into the SQL string. A value containing a single quote (e.g. \`acme's-org\`) will produce a syntax error; a crafted value can inject arbitrary SQL. \`execute_unprepared\` is intended for DDL statements that cannot be parameterized, not for DML with user-controlled values.

Fix using SeaORM's parameterized \`execute\`:

\`\`\`rust
use sea_orm::{Statement, Value};

async fn update_null_org_rows(
    db: &sea_orm::DatabaseConnection,
    table: &str,
    org_id: &str,
) -> Result<u64, sea_orm::DbErr> {
    let stmt = Statement::from_sql_and_values(
        db.get_database_backend(),
        format!("UPDATE {} SET organization_id = ? WHERE organization_id IS NULL", table),
        vec![Value::from(org_id.to_owned())],
    );
    let result = db.execute(stmt).await?;
    Ok(result.rows_affected())
}
\`\`\`

The \`table\` identifier comes from the hardcoded \`SCOPED_TABLES\` constant, not from user input, so that interpolation is acceptable -- a brief comment noting this assumption would help future readers.

---

### Non-blocking suggestions

**1. Hardcoded \`DatabaseBackend::Sqlite\` in \`count_null_org_rows\`** (line ~95)

\`\`\`rust
sea_orm::Statement::from_string(
    sea_orm::DatabaseBackend::Sqlite,
    ...
)
\`\`\`

The live connection knows its own backend. Prefer \`db.get_database_backend()\` -- same fix shown in the snippet above. Costs nothing and survives a future backend change without silent breakage.

---

**2. Missing input validation for \`--org-id\`** (admin.rs ~line 130)

An empty \`--org-id ""\` would silently set \`organization_id = ''\` on all rows -- not a crash, but wrong. Fail early:

\`\`\`rust
if org_id.is_empty() {
    anyhow::bail!("--org-id must not be empty");
}
\`\`\`

---

**3. No tests**

The \`--dry-run\` path is the key safety mechanism and is entirely untested. The common storage module exposes \`create_test_connection\` for in-memory SQLite. A minimal test should:
- Seed rows with \`organization_id = NULL\`
- Assert \`--dry-run\` returns the correct count and does **not** mutate data
- Assert the live update assigns \`org_id\` correctly and returns the right affected count

---

**4. Transition-period security implication understated in \`tenant.rs\` docs**

The NULL-handling policy ("return rows where \`organization_id\` matches OR is NULL") means **every authenticated tenant can read all legacy unscoped data** during the transition window. This is intentional but the current docs don't make the scope of exposure clear. Suggest adding an explicit callout:

\`\`\`
/// # Security note
///
/// During the transition window, all tenants share read access to
/// NULL-scoped (legacy) rows. Run \`agent admin backfill-tenant\` promptly
/// after deploying the migration to close this window.
\`\`\`

---

### What's well done

- \`--dry-run\` mode is implemented as a first-class path, not an afterthought.
- Skipping databases that don't exist on disk is graceful and operator-friendly.
- JSON output mode is consistent with the rest of the CLI.
- Validating service names against \`DB_SERVICES\` before opening connections is good defensive practice.
- The doc-comment on \`BackfillTenant\` accurately describes when and why to use the command.

Please fix the SQL injection before merge. Items 1-4 above are noted for follow-up.

PR #1291, opened 2026-06-14
feat(ui): simplify proxy to single gateway route

### LINKED ISSUE #1121: Auth cleanup: UI proxy simplification, inter-service tenant propagation, MCP docs
## Context

After auth is integrated across the CLI, UI, and services, several cleanup tasks remain: the UI proxy service becomes redundant for API calls, inter-service calls need to propagate tenant context, and the MCP server trust model should be documented.

## Checklist

- [ ] **Simplify UI proxy service** -- Once the React app routes through the core gateway, the `crates/ui` API proxy is redundant for API requests. Simplify it to only serve static files, or update it to forward API requests to the core gateway instead of individual services.
  - Files: `crates/ui/src/proxy.rs`, `crates/ui/src/lib.rs`, `crates/ui/src/config.rs`

- [ ] **Inter-service tenant propagation** -- Services that call other services (orchestrator calls communicate for room joins, message bridge relays messages) need to propagate `X-Tenant-ID`. Add `with_tenant_id()` to service clients used for inter-service calls so the header is injected on every request.
  - Files: `crates/orchestrator/src/main.rs`, `crates/orchestrator/src/message_bridge.rs`, `crates/communicate/src/client.rs`

- [ ] **Document MCP trust model** -- MCP server stays local/trusted. Document that it works because `OptionalTenantId` returns None and queries return unscoped results in permissive mode. No functional changes needed.
  - Files: `crates/mcp/src/client.rs` (doc comment)

## Dependencies

- UI auth integration issue (gateway routing must be in place before simplifying the proxy)
- Service tenant scoping issue (inter-service propagation depends on tenant-scoped services)

geoffjay:
Replace the three per-service proxy handlers (`proxy_ask`, `proxy_notify`, `proxy_orchestrator`) with a single `proxy_to_gateway` handler that forwards all `/api/**` requests to the core gateway at `/api/v1/**`. `ProxyState` now holds a single `gateway_url` instead of three per-service URLs, and the UI is configured via `AGENTD_CORE_SERVICE_URL` instead of the three separate service-specific env vars.

Closes #1121

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,7 @@ dependencies = [
  "hyper",
  "metrics",
  "metrics-exporter-prometheus",
+ "notify",
  "rand 0.8.5",
  "reqwest",
  "sea-orm",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -38,3 +38,4 @@ tower = { version = "0.5", features = ["util"] }
 http-body-util = "0.1"
 hyper = "1.0"
 tempfile = "3.14"
+notify = { path = "../notify" }
--- a/crates/core/tests/e2e_auth_flow.rs
+++ b/crates/core/tests/e2e_auth_flow.rs
@@ -1,0 +1,203 @@
+//! End-to-end authentication and tenant-isolation integration test.
+//!
+//! This test exercises the complete auth chain in a single process, with no
+//! external binaries or persistent databases:
+//!
+//! ```text
+//!  [test]
+//!    |
+//!    |-- POST /auth/register          --> core (in-process listener A)
+//!    |<-- 201 { token, user.active_organization_id }
+//!    |
+//!    |-- POST /api/v1/notify/notifications  Bearer <token_a>
+//!    |   core gateway: reads token -> resolves org -> injects X-Tenant-ID
+//!    |   --> notify (in-process listener B): stores notification with org_id
+//!    |<-- 201
+//!    |
+//!    |-- GET /api/v1/notify/notifications   Bearer <token_a>
+//!    |<-- 200 { total: 1 }   (user A sees their own notification)
+//!    |
+//!    |-- POST /auth/register          (user B, separate org)
+//!    |-- GET /api/v1/notify/notifications   Bearer <token_b>
+//!    |<-- 200 { total: 0 }   (user B cannot see user A's notification)
+//! ```
+//!
+//! Acceptance criteria (from issue #1128):
+//! - [x] register -> login -> gateway proxy -> downstream with X-Tenant-ID
+//! - [x] tenant isolation: user A's data is invisible to user B
+//! - [x] temp databases — no side effects on dev data
+//! - [x] passes with `cargo test`
+
+use agentd_common::storage::create_test_connection;
+use agentd_core::{
+    api::{create_router_with_proxy, AppState},
+    proxy::ProxyConfig,
+    storage::Storage,
+};
+use notify::{
+    api::{create_router as notify_create_router, ApiState as NotifyState},
+    storage::NotificationStorage,
+};
+use reqwest::Client;
+use serde_json::{json, Value};
+use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/// Bind a `TcpListener` on an OS-assigned ephemeral port and return its address.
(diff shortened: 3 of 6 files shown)


geoffjay (commented):
## Review: feat(ui): simplify proxy to single gateway route

### Stack context

PR #1291 is the **top** of the stack: #1290 (backfill, needs-rework) → #1289 (tenant scoping, needs-rework) → #1288 (UI auth, needs-rework) → #1287 (CLI gateway routing, merge-queue) → #1129 (base, open). This PR cannot merge until all parents land, but it has no blocking issues of its own.

---

The simplification is correct and well-motivated. The path rewriting logic handles the important cases:

```
/api/notify/notifications  →  strip "/api"  →  "/notify/notifications"
                           →  add  "/api/v1" →  http://gateway/api/v1/notify/notifications  ✓

/api/ask/questions         →  http://gateway/api/v1/ask/questions  ✓
/api                       →  "" tail        →  http://gateway/api/v1               ✓
```

The `unwrap_or("")` fallback is safe: the Axum route `/api/{*path}` always matches with `/api` as a prefix, so `strip_prefix("/api")` will never return `None` in practice.

Net diff is -21 lines. Three dead functions removed. The change is well-scoped.

---

### Non-blocking suggestions

**1. Previously-missing services are now implicitly proxied — worth a comment**

The old proxy only covered `/api/ask/**`, `/api/notify/**`, `/api/orchestrator/**`. Memory, communicate, monitor, and wrap had no proxy routes. The new `/api/{*path}` wildcard silently picks all of them up. This is almost certainly the desired behavior, but a note in the `proxy_to_gateway` doc-comment listing the covered services (or explicitly saying "all `/api/**` traffic") would make it obvious to future readers that the broadened coverage is intentional, not accidental.

---

**2. `runtime_config.rs` does not expose `core` to the SPA**

`BROWSER_SERVICES` intentionally excludes `core`:

```rust
// hook, wrap, and core are backend-only and intentionally not exposed.
const BROWSER_SERVICES: &[&str] =
    &["ask", "notify", "orchestrator", "memory", "monitor", "communicate"];
```

But the SPA code added in parent PR #1288 uses `runtimeServiceUrl("core")` in both `services/config.ts` and `services/auth.ts`. Because `core` is absent from the runtime config document, `runtimeServiceUrl("core")` always resolves to `null`, so the SPA falls back to the hardcoded `"http://localhost:17000"`. Auth calls and all service calls in #1288 bypass this proxy entirely — they go direct to `localhost:17000` rather than through `/api/**`.

This is a stack coherence issue: either
- #1288 should use relative proxy paths (`/api/notify/...`) instead of absolute service URLs, and `runtimeServiceUrl("core")` should be removed from the SPA, or
- `core` should be added to `BROWSER_SERVICES` so the SPA can locate the gateway dynamically in non-localhost deployments.

Neither change belongs in this PR, but the gap should be resolved in the #1288 rework before the stack merges. Flagging here so it isn't missed.

---

**3. Hop-by-hop headers are forwarded as-is**

```rust
for (key, value) in headers.iter() {
    if key != "host" {
        upstream_req = upstream_req.header(key, value);
    }
}
```

Headers like `connection`, `keep-alive`, `transfer-encoding`, `te`, `trailer`, and `upgrade` are hop-by-hop and should be stripped before forwarding (RFC 7230 §6.1). Forwarding them can cause the upstream to misinterpret the body framing or the connection lifecycle. This is **pre-existing behavior** and not introduced by this PR, but a follow-up ticket to strip hop-by-hop headers would make this a more correct HTTP proxy. Low priority for an internal proxy over loopback.

---

**4. No tests for the new path-rewriting logic**

`proxy_to_gateway` and the updated two-argument `proxy_request` have no test coverage. A unit test that constructs a mock request to `/api/notify/foo?bar=1` and asserts the forwarded URL is `http://gateway/api/v1/notify/foo?bar=1` would lock in the rewriting contract and catch regressions if the prefix constants are changed. This was also absent before this PR, so not a regression.

---

### What's well done

- The simplification eliminates the state bloat (`ask_url`, `notify_url`, `orchestrator_url`) and the three near-identical handler functions.
- Adding `strip_prefix`/`add_prefix` parameters to `proxy_request` makes the rewriting logic explicit and reusable rather than hardcoded.
- The doc-comment example (`GET /api/notify/notifications → GET {gateway}/api/v1/notify/notifications`) is exactly the right level of documentation for a proxy function.
- Config, lib, and proxy are all updated consistently — no stale references to the old per-service fields.
- Removing the three separate `AGENTD_*_SERVICE_URL` env vars from the UI config is consistent with the gateway-first direction established in #1287.

PR #1292, opened 2026-06-14
test(core): E2E auth flow and cross-tenant isolation test

### LINKED ISSUE #1128: End-to-end auth integration test across CLI, gateway, and downstream services
## Context

After the core service auth integration is complete (#1116-#1121), there is no automated test that validates the full chain: CLI login -> token storage -> CLI command routed through core gateway -> downstream service receives X-Tenant-ID and scopes data correctly.

Each individual issue adds unit/integration tests for its own scope, but no test exercises the complete flow end-to-end. This gap means regressions in the auth chain could go undetected.

## Implementation

Add an integration test (or small test binary) that exercises the full auth flow:

### Test scenario

1. **Start services**: Launch agentd-core and at least one downstream service (e.g., notify) in test mode with temp databases
2. **Register a user**: POST /auth/register to core, receive a session token
3. **Create an organization**: POST /api/v1/organizations via the gateway
4. **Route through gateway**: Create a notification via the gateway path (POST /api/v1/notify/notifications) with the bearer token
5. **Verify tenant scoping**: The notification should have the user's active organization_id
6. **Verify cross-tenant isolation**: Create a second user/org, verify they cannot see the first user's notification

### Test approach options

**Option A: In-process test** (preferred if feasible)
- Compose the core router and a downstream router in the same test process
- Use `axum::Router::merge` or wire the proxy to an in-memory service
- Avoids spawning real processes

**Option B: Process-based test**
- Use `cargo build` artifacts and spawn services on ephemeral ports
- More realistic but slower and flakier

### Files

- `crates/core/tests/e2e_auth_flow.rs` (new) - or a dedicated `tests/` workspace crate if cross-crate

## Acceptance Criteria

- [ ] Test exercises: register -> login -> gateway proxy -> downstream service with X-Tenant-ID
- [ ] Test verifies tenant isolation (user A cannot see user B's data)
- [ ] Test uses temp databases (no side effects on dev data)
- [ ] Test passes in CI (`cargo test`)
- [ ] Documents the expected auth flow in test comments for onboarding reference

## Relevant Files

- `crates/core/src/api/gateway.rs` - gateway proxy handler
- `crates/core/src/middleware/tenant.rs` - TenantContext extractor
- `crates/core/src/proxy.rs` - proxy_request with X-Tenant-ID injection
- `crates/common/src/tenant.rs` - downstream tenant extractor (from #1116)

## Blocked By
- #1117 (CLI auth integration must be complete)
- #1119 (downstream services must have tenant scoping)

## Stack Base
Branch off `feature/autonomous-pipeline`.
Blocked by: #1117, #1119

geoffjay:
Adds `crates/core/tests/e2e_auth_flow.rs` exercising the complete auth chain in a single process with no external binaries or persistent databases:

- Two in-process services (core + notify) are composed on ephemeral TCP listeners
- `POST /auth/register` creates a user with a personal org and returns a session token
- `POST /api/v1/notify/notifications` with the bearer token routes through the gateway, which resolves the active org and injects `X-Tenant-ID` before forwarding to notify
- The notify service stores `organization_id` on the notification row (from `X-Tenant-ID`)
- A second user registers with a separate org and receives zero results on list — confirming cross-tenant isolation
- `notify` added as a dev-dependency of `agentd-core` for the test

Closes #1128

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,7 @@ dependencies = [
  "hyper",
  "metrics",
  "metrics-exporter-prometheus",
+ "notify",
  "rand 0.8.5",
  "reqwest",
  "sea-orm",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -38,3 +38,4 @@ tower = { version = "0.5", features = ["util"] }
 http-body-util = "0.1"
 hyper = "1.0"
 tempfile = "3.14"
+notify = { path = "../notify" }
--- a/crates/core/tests/e2e_auth_flow.rs
+++ b/crates/core/tests/e2e_auth_flow.rs
@@ -1,0 +1,203 @@
+//! End-to-end authentication and tenant-isolation integration test.
+//!
+//! This test exercises the complete auth chain in a single process, with no
+//! external binaries or persistent databases:
+//!
+//! ```text
+//!  [test]
+//!    |
+//!    |-- POST /auth/register          --> core (in-process listener A)
+//!    |<-- 201 { token, user.active_organization_id }
+//!    |
+//!    |-- POST /api/v1/notify/notifications  Bearer <token_a>
+//!    |   core gateway: reads token -> resolves org -> injects X-Tenant-ID
+//!    |   --> notify (in-process listener B): stores notification with org_id
+//!    |<-- 201
+//!    |
+//!    |-- GET /api/v1/notify/notifications   Bearer <token_a>
+//!    |<-- 200 { total: 1 }   (user A sees their own notification)
+//!    |
+//!    |-- POST /auth/register          (user B, separate org)
+//!    |-- GET /api/v1/notify/notifications   Bearer <token_b>
+//!    |<-- 200 { total: 0 }   (user B cannot see user A's notification)
+//! ```
+//!
+//! Acceptance criteria (from issue #1128):
+//! - [x] register -> login -> gateway proxy -> downstream with X-Tenant-ID
+//! - [x] tenant isolation: user A's data is invisible to user B
+//! - [x] temp databases — no side effects on dev data
+//! - [x] passes with `cargo test`
+
+use agentd_common::storage::create_test_connection;
+use agentd_core::{
+    api::{create_router_with_proxy, AppState},
+    proxy::ProxyConfig,
+    storage::Storage,
+};
+use notify::{
+    api::{create_router as notify_create_router, ApiState as NotifyState},
+    storage::NotificationStorage,
+};
+use reqwest::Client;
+use serde_json::{json, Value};
+use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/// Bind a `TcpListener` on an OS-assigned ephemeral port and return its address.
+async fn bind_ephemeral() -> (SocketAddr, tokio::net::TcpListener) {
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr = listener.local_addr().unwrap();
+    (addr, listener)
+}
+
+/// Start an in-process notify service backed by a temp SQLite file.
+///
+/// Returns the bound address and the `TempDir` guard — the caller must keep
+/// the guard alive for the duration of the test.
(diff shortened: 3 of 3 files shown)


geoffjay (commented):
## Review: test(core): E2E auth flow and cross-tenant isolation test

### Stack context

This PR sits at the **top** of the stack: #1291 (proxy simplification, merge-queue) → #1290 (backfill, needs-rework) → #1289 (tenant scoping, needs-rework) → #1288 (UI auth, needs-rework) → #1287 (gateway routing, merge-queue) → #1129 (auth completeness, open).

Because three parent PRs carry `needs-rework`, this test cannot run green until those fixes land. Importantly, the test's assertions are written for the **post-fix** behavior of #1289 (OR-NULL filter, org_id stored on create) — a deliberate choice that makes sense for an acceptance test living at the top of the stack.

---

All the structural concerns check out:

- `notify` is correctly in `[dev-dependencies]`, not `[dependencies]` — no production binary bloat.
- `NotificationStorage::with_path` exists (line 102 of `notify/src/storage.rs`).
- `create_router_with_proxy` and `AppState` are public in `crates/core/src/api.rs` (line 46).
- `ProxyConfig` is exported from `crates/core::proxy`.
- Both `TempDir` guards are kept in scope correctly — `notify_tmp` (bound, not dropped) and `_core_tmp` (underscore retains the value without triggering the unused-variable lint).
- Ephemeral `127.0.0.1:0` ports eliminate any risk of port conflicts in CI.
- The spawned server tasks can handle requests when the test body reaches any `.await` point under the default single-threaded Tokio executor — no `multi_thread` flavor needed here.

The test correctly asserts both the positive case (Alice sees 1) and the isolation case (Bob sees 0), and the acceptance criteria checklist in the module doc matches issue #1128.

---

### Non-blocking suggestions

**1. OR-NULL comment is slightly ahead of the implementation**

The inline comments describe the post-fix filter behavior:

```rust
// the tenant-scoped list (org_id = org_id_a OR org_id IS NULL) returns it.
```

and:

```rust
// Bob's query filters to org_id_b OR NULL.
```

#1289 currently uses strict equality, not OR-NULL (that fix is in its `needs-rework`). The comments correctly anticipate the target behavior, so they will be accurate once #1289 is fixed — no change needed unless the OR-NULL fix is descoped from #1289. Worth a note here so the #1289 author knows these comments are the acceptance contract they're writing toward.

---

**2. The OR-NULL transition-period behavior is not exercised**

The test proves that a scoped notification (`organization_id = org_id_a`) is invisible to other tenants. It does not verify the other half of the policy documented in `tenant.rs`: that a NULL-scoped legacy notification is visible to **all** authenticated tenants (the transition-period guarantee).

A complementary test case would:
1. Insert a notification directly into the notify DB with `organization_id = NULL` (bypassing the gateway)
2. Assert that both Alice and Bob can see it when listing

This is the scenario that breaks when the OR-NULL filter is accidentally tightened to strict equality. As written, the existing test would still pass under strict equality, so it does not guard against that regression. Consider adding it as a second `#[tokio::test]` in the same file, or as additional steps after step 6.

---

**3. Spawned server task panics are silent**

```rust
tokio::spawn(async move { axum::serve(listener, router).await.unwrap() });
```

If the server task panics after startup (e.g., a request handler panics), the panic is swallowed by Tokio and the test fails with a confusing "connection refused" or garbled response rather than the real error. For a test this simple the risk is low, but `axum::serve(...).await.expect("notify server crashed")` is no worse and gives a clearer failure message in CI.

---

**4. No assertion on the created notification's `organization_id`**

The create step only asserts the HTTP status:

```rust
assert_eq!(create_status.as_u16(), 201, "notification creation should return 201");
```

Fetching the created notification's ID from the response and asserting its `organization_id == org_id_a` would directly verify that the gateway injected `X-Tenant-ID` correctly and that the notify service persisted it — the exact mechanism the test description highlights. This would make the test a more precise regression guard for the header-injection path.

---

### What's well done

- The module-level doc with the ASCII diagram and the acceptance-criteria checklist is exactly the right amount of context for a test this complex.
- Composing two real in-process services (core + notify) against ephemeral SQLite files and ephemeral TCP ports is a clean, hermetic design that avoids any flakiness from shared state.
- The `register()` helper extracts both token and org_id in one call, making the test body readable.
- `assert_ne!(org_id_b, org_id_a)` before the isolation assertion is a nice guard that would catch a hypothetical bug where all users share the same org.

PR #1295, opened 2026-06-16
fix(ui): expose core service in runtime /config.json

geoffjay:
The SPA calls the core service directly for auth (/auth/register,
/auth/login) via `runtimeServiceUrl("core")`, but `core` was omitted
from BROWSER_SERVICES, so /config.json never carried it. The frontend
fell through to the compiled-in `http://localhost:17000` default,
ignoring AGENTD_CORE_SERVICE_URL (backend-only) and the build-time
VITE_AGENTD_CORE_SERVICE_URL (baked into the prebuilt bundle). On a
remote host this produced ERR_CONNECTION_REFUSED against localhost.

Add `core` to BROWSER_SERVICES and its port lookup so the SPA derives
the core URL from the page host + port, and honors a
[services.ui.public_urls] core override. Document the core override in
install.md and add test coverage.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/ui/src/runtime_config.rs
+++ b/crates/ui/src/runtime_config.rs
@@ -29,10 +29,11 @@ pub struct ServiceEntry {
     pub url: Option<String>,
 }
 
-/// Services the SPA talks to directly. `hook`, `wrap`, and `core` are
-/// backend-only and intentionally not exposed.
+/// Services the SPA talks to directly. The browser calls `core` for auth
+/// (`/auth/register`, `/auth/login`), so it must be exposed here; `hook` and
+/// `wrap` are backend-only and intentionally not exposed.
 const BROWSER_SERVICES: &[&str] =
-    &["ask", "notify", "orchestrator", "memory", "monitor", "communicate"];
+    &["core", "ask", "notify", "orchestrator", "memory", "monitor", "communicate"];
 
 /// Build the runtime configuration from the shared agentd config.
 pub fn build(shared: &agentd_common::config::AgentdConfig) -> RuntimeConfig {
@@ -42,6 +43,7 @@ pub fn build(shared: &agentd_common::config::AgentdConfig) -> RuntimeConfig {
     let mut services = BTreeMap::new();
     for &name in BROWSER_SERVICES {
         let port = match name {
+            "core" => s.core.port,
             "ask" => s.ask.port,
             "notify" => s.notify.port,
             "orchestrator" => s.orchestrator.port,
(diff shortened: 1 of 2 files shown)


geoffjay (commented):
## Review: fix(ui): expose core service in runtime /config.json

**Verdict: Approved ✅** *(self-review — cannot use the approve action on own PR)*

### Stack Position

This branch sits directly on `feature/core-service-url`, which is a long-lived feature scope branch without its own PR (expected per project workflow). No parent PR needs review first; the base branch exists on the remote and is the merge target.

### Root Cause — Confirmed Correct

The diagnosis is accurate. `runtimeServiceUrl("core")` was returning `undefined` because `core` was absent from the serialized `/config.json`, so `serviceConfig.coreServiceUrl` in `ui/src/services/config.ts` fell through all three resolution tiers straight to the hard-coded `"http://localhost:17000"` default — completely ignoring the runtime host. The fix is applied at the right layer.

### Code Changes

**`crates/ui/src/runtime_config.rs`**

- Adding `"core"` to `BROWSER_SERVICES` and the corresponding `"core" => s.core.port` match arm is the minimal, correct change. The `_ => unreachable!()` guard is preserved — any future service added to the slice that misses a port arm will panic immediately in tests rather than silently slipping through.
- Port default 17000 verified against `CoreConfig::default()` in `crates/common/src/config.rs` (line 340 confirms `Self { port: 17000 }`).
- The updated doc comment on `BROWSER_SERVICES` accurately reflects the new split: hook/wrap are still backend-only; core is now correctly described as browser-facing.

**Tests**

- `test_build_default_ports`: the added `assert_eq!(runtime.services["core"].port, 17000)` assertion makes a future accidental revert fail loudly.
- `test_core_public_url_override`: correctly exercises the `public_urls` override path for `core`, consistent with the existing `test_build_with_public_url_override` pattern for orchestrator. Clean and sufficient.

**`docs/public/install.md`**

The note explaining that *core* handles login/registration gives operators the WHY, not just the WHAT. The added TOML example places the override in the correct section.

### Minor Suggestion (non-blocking)

The frontend test fixture in `ui/src/test/runtime-config.test.ts` uses only `ask` and `orchestrator` entries — it would be good to add `core` and a dedicated `runtimeServiceUrl("core")` case in a follow-up to keep TS-layer coverage symmetric with the new Rust tests. Not a blocker: the function is generic and the port-derive + url-override paths are already well covered.

### Checklist

- [x] Correctness: root cause correctly identified and fixed
- [x] Safety: no new unwraps, no new panic surfaces
- [x] Consistency: follows the exact same port-match pattern as every other service
- [x] Tests: default-port assertion added; dedicated public-URL override test added
- [x] Scope: 2 files, tightly focused on the stated fix — no unrelated changes
- [x] Docs: `install.md` updated with the core override example

PR #1297, opened 2026-06-16
feat(knowledge): KB-1 — crate scaffold + KnowledgeConfig + service registration

geoffjay:
Scaffolds the new `agentd-knowledge` service crate with config, migration, health endpoint, and full service registration across the workspace.

Closes #1280

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,6 +74,7 @@ dependencies = [
  "anyhow",
  "colored 2.2.0",
  "communicate",
+ "knowledge",
  "memory-api",
  "notify",
  "orchestrator",
@@ -1284,6 +1285,16 @@ name = "constant_time_eq"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
+
+[[package]]
+name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
 
 [[package]]
 name = "core-foundation"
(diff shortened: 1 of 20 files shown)


geoffjay (commented):
## Review: KB-1 — crate scaffold + KnowledgeConfig + service registration

**Approved** — the scaffold is well-structured and all integration points are wired correctly. A few non-blocking suggestions below.

---

### Stack note

This PR is stacked on #1296 (the `epic-1279` branch). Neither can merge until #1296 merges first, but the changes here are self-contained and reviewable independently.

---

### What looks good

- **Config integration** follows the exact `ServicesConfig` / `merge()` / `apply_env_overrides()` pattern used by every other service — complete and correct.
- **Port allocation** — 17011 is unassigned; no conflict with existing services.
- **Migration is idempotent** — all `create_table` / `create_index` calls use `if_not_exists()`. The `down()` only drops the table; SQLite drops its indexes automatically, so this is correct.
- **Error types** correctly use `thiserror` with `#[from] anyhow::Error` for the `Other` variant, and `anyhow` for the binary entry-point — matches the project's `thiserror`-for-libs / `anyhow`-for-binaries convention.
- **Xtask registration** is complete: `DB_SERVICES`, `migrate()` arm, `migrate_status()` arm, `platform::SERVICES`, `SERVICE_NAMES`, and the completeness test counter are all updated.
- **xtask port** — `7011` in `platform::SERVICES` is correct; every other service follows the `70XX` dev-port scheme (the workspace config defaults use `170XX`).
- **`ValidateConfig` for `KnowledgeConfig`** in common/config.rs delegates to `validate_port()` — matches every other service's impl.

---

### Non-blocking suggestions

**`crates/knowledge/Cargo.toml` — `reqwest` should use `workspace = true`**

```toml
# current
reqwest = { version = "0.12", features = ["json"] }

# suggested
reqwest = { workspace = true, features = ["json"] }
```

The workspace definition has `default-features = false, features = ["rustls-tls"]`. By specifying the dep directly without `default-features = false`, Cargo will compile reqwest with its default feature set enabled (which may include `native-tls`), diverging from every other crate in the workspace. `rustls-tls` is also missing. Cargo merges features across usages so this won't cause a build error, but it can silently pull in a C TLS dependency. Worth fixing before KB-4 (client implementation).

---

**`crates/knowledge/Cargo.toml` — `tower` in `[dependencies]` appears unused in production code**

```toml
# in [dependencies]
tower = "0.5"                                    # no direct tower:: usage in production code
# in [dev-dependencies]
tower = { version = "0.5", features = ["util"] } # needed for ServiceExt in tests
```

No production file calls `tower::` directly; `axum` and `tower-http` cover the server-side needs. Moving it to `[dev-dependencies]` only (with the `util` feature) would be cleaner.

---

**`crates/knowledge/Cargo.toml` — `directories` may be an unneeded direct dep**

```toml
directories = "6.0.0"
```

The only `ProjectDirs` usage for this service (`default_knowledge_docs_path()`) lives in `agentd-common/src/config.rs`, not in the knowledge crate itself. If KB-2 won't call `directories` directly, this can be dropped.

---

**`crates/knowledge/src/api.rs` — dead `create_router()` stub**

```rust
#[allow(dead_code)]
pub fn create_router() -> Router {
    Router::new().route("/health", get(health_handler))
}
```

`create_router_with_state()` is the only used constructor. If this no-state variant isn't needed for upcoming tests, removing it avoids confusion about which builder to reach for. If it _is_ a future test convenience, a short doc comment explaining that would help.

---

All other aspects — types, entity model, migration, storage stub, client stub, UI config wiring — are clean and consistent with the rest of the codebase. Ready to proceed once the `reqwest` workspace dep is noted for KB-4.

PR #1300, opened 2026-06-16
feat(knowledge): KB-4 client and CLI commands

geoffjay:
Add \`KnowledgeClient\` HTTP facade and \`agent knowledge\` CLI subcommands for the agentd-knowledge service.

**KnowledgeClient** (\`crates/knowledge/src/client.rs\`):
- Typed methods for all REST endpoints: \`health\`, \`list_documents\`, \`create_document\`, \`bulk_delete_documents\`, \`get_document\`, \`get_document_content\`, \`update_document\`, \`delete_document\`, \`get_tree\`
- \`from_env()\` reads \`AGENTD_KNOWLEDGE_SERVICE_URL\` or falls back to shared config port (17011)
- \`with_token()\` attaches a bearer token for core gateway routing
- All methods now check HTTP status before parsing JSON (blocking review fix)
- Inline URL encoding helper for query params; path segments assumed UUID (documented with TODO)

**CLI subcommands** (\`crates/cli/src/commands/knowledge.rs\`):
- \`agent knowledge health\` — service health check
- \`agent knowledge list <project_id>\` — paginated listing with optional \`--prefix\` filter
- \`agent knowledge get <project_id> <doc_id>\` — document metadata
- \`agent knowledge content <project_id> <doc_id>\` — metadata + markdown body
- \`agent knowledge create <project_id> <rel_path>\` — inline \`--content\` or \`--from-file\`
- \`agent knowledge update <project_id> <doc_id>\` — update content/title with optional optimistic concurrency \`--expected-updated-at\`
- \`agent knowledge delete <project_id> <doc_id>\` — delete single document
- \`agent knowledge gc <project_id> --yes\` — bulk-delete all project documents (requires \`--yes\` guard)
- \`agent knowledge tree <project_id>\` — display virtual folder/file tree
- All string slice indexing uses \`.get(..N).unwrap_or()\` to avoid panics on short strings (blocking review fix)
- All commands support \`--json\` for raw JSON output

**Review feedback addressed:**
1. HTTP status checks added to \`health\`, \`list_documents\`, \`get_tree\` (were calling \`.json()\` directly)
2. Unchecked string slice panics fixed in \`cmd_list\`, \`cmd_get\` (use \`.get(..N).unwrap_or()\`)
3. \`gc\` command requires \`--yes\` flag to prevent accidental bulk deletion
4. \`#![allow(dead_code)]\` removed from \`client.rs\`; \`mod client\` removed from binary \`main.rs\` (client lives in lib only)
5. \`orchestrator.rs\` whitespace-only diff reverted
6. \`cmd_update\` refactored to stay within 7-argument clippy limit

**Tests**: 3 unit tests for \`resolve_content\`; 37+41 knowledge crate tests all pass.

Closes #1283


crates/knowledge/src/client.rs
```diff
@@ -43,4 +55,198 @@ impl KnowledgeClient {
...
     }
+
+    // -----------------------------------------------------------------------
+    // Documents — collection
+    // -----------------------------------------------------------------------
+
+    /// List documents for `project_id` with optional prefix filter and pagination.
+    pub async fn list_documents(
+        &self,
+        project_id: &str,
```
geoffjay, line 89: **Blocking:** `list_documents` does not call `is_success()` before `.json()`. A 404 or 500 from the server produces `"failed to parse list_documents response: expected value at line 1 column 1"` — misleading noise. All mutating methods and every GET in `communicate/src/client.rs` use the status-check pattern. `get_tree()` has the same gap.

crates/cli/src/commands/knowledge.rs
```diff
@@ -0,0 +1,502 @@
...
+                    doc_id,
+                    title.as_deref(),
+                    content.as_deref(),
+                    from_file.as_ref(),
+                    expected_updated_at.as_deref(),
+                    json,
+                )
+                .await
+            }
+            KnowledgeCommand::Delete { project_id, doc_id } => {
```
geoffjay, line 270: **Blocking:** `&doc.id[..8]` panics if the server returns an ID shorter than 8 bytes. `id` is `String` in the entity model — no UUID type guarantee. Use `doc.id.get(..8).unwrap_or(&doc.id)` to avoid a CLI crash on unexpected server data.

crates/cli/src/commands/knowledge.rs
```diff
@@ -0,0 +1,502 @@
...
+                cmd_delete(client, project_id, doc_id, json).await
+            }
+            KnowledgeCommand::Gc { project_id } => cmd_gc(client, project_id, json).await,
+            KnowledgeCommand::Tree { project_id } => cmd_tree(client, project_id, json).await,
+        }
+    }
```
geoffjay, line 281: **Blocking:** `&doc.updated_at[..19]` panics if the timestamp is shorter than 19 bytes. Use `doc.updated_at.get(..19).unwrap_or(&doc.updated_at)`. Same fix needed for `created_at[..19]` in `cmd_get` (~line 273) and `cmd_content` (~line 292).

crates/cli/src/commands/knowledge.rs
```diff
@@ -0,0 +1,502 @@
...
+    expected_updated_at: Option<&str>,
+    json: bool,
+) -> Result<()> {
+    let body = if content.is_some() || from_file.is_some() {
+        Some(resolve_content(content, from_file)?)
+    } else {
+        None
+    };
+    let req = UpdateDocumentRequest {
+        content: body,
```
geoffjay, line 417: **Should fix:** `gc` permanently deletes all project documents with no confirmation gate. Project precedent: `config init --force` requires an explicit flag before overwriting. Add `yes: bool` to `Gc` and bail with a clear message unless `--yes` is passed.

geoffjay (commented):
## Review: KB-4 client and CLI commands

**Changes requested** (owner PR — cannot formally request, but these should block merge) — two blocking bugs plus a significant UX gap.

---

### Stack note

This PR is stacked on #1299 (KB-3 REST API layer → `issue-1282`). It cannot merge until KB-3 lands. The implementation here is independent and reviewable now.

---

### Blocking issues

#### 1. Missing HTTP status checks on `list_documents` and `get_tree`

`health()`, `list_documents()`, and `get_tree()` call `.json()` directly without first checking `is_success()`. The other eight methods in the same file all follow the correct guard pattern:

```rust
if res.status().is_success() {
    res.json().await.context("…")
} else {
    let status = res.status();
    let body: Value = res.json().await.unwrap_or(Value::Null);
    anyhow::bail!("… failed ({status}): {body}");
}
```

The `communicate` client (the project's closest analog — `crates/communicate/src/client.rs:418-430`) checks status on every GET. Without the guard, a 404 or 503 from the server silently turns into `"failed to parse list_documents response: expected value at line 1 column 1"` — a confusing, misleading error.

**Files / approximate lines:**
- `crates/knowledge/src/client.rs` — `list_documents` (the `self.client.get(&url)…json()` chain, ~line 66 of the method body)
- `crates/knowledge/src/client.rs` — `get_tree` (same pattern at the end of the file)

#### 2. Unchecked string-slice panics in CLI display code

`cmd_list`, `cmd_get`, and `cmd_content` all index server-controlled strings without bounds checks:

```rust
let short_id = &doc.id[..8];           // panics if id is < 8 bytes
Cell::new(&doc.updated_at[..19]);       // panics if timestamp is < 19 bytes
&doc.created_at[..19]                   // same
```

The entity model stores `id: String` — no validated UUID type, no minimum-length guarantee. If the service ever returns a non-UUID ID or a truncated timestamp the CLI will panic rather than show a clean error. Safe drop-in replacements:

```rust
let short_id = doc.id.get(..8).unwrap_or(&doc.id);
let updated  = doc.updated_at.get(..19).unwrap_or(&doc.updated_at);
```

**Files / approximate lines:**
- `crates/cli/src/commands/knowledge.rs` — `cmd_list` (~line 239, `let short_id = &doc.id[..8]`)
- `crates/cli/src/commands/knowledge.rs` — `cmd_list` (~line 245, `&doc.updated_at[..19]`)
- `crates/cli/src/commands/knowledge.rs` — `cmd_get`, `cmd_content` (both `created_at[..19]` and `updated_at[..19]`)

---

### Significant suggestion — `gc` needs a `--yes` / `--force` guard

The doc comment says "This is irreversible" yet the command runs with no confirmation. The established project precedent is a `--force` flag for destructive ops (see `agent config init --force`, `crates/cli/src/commands/config.rs:43–52`). Add a `yes: bool` arg to `Gc` and gate the call:

```rust
Gc {
    project_id: String,
    /// Skip the confirmation prompt. Required for non-interactive use.
    #[arg(long)]
    yes: bool,
},
```

```rust
async fn cmd_gc(client: &KnowledgeClient, project_id: &str, yes: bool, json: bool) -> Result<()> {
    if !yes {
        anyhow::bail!("This will permanently delete all documents for project {project_id}. Pass --yes to confirm.");
    }
    …
}
```

---

### Minor / non-blocking

- **`#![allow(dead_code)]`** at top of `client.rs` — the client is now fully implemented and used; this suppressor can be removed.
- **`orchestrator.rs` diff** — line 1925 is a whitespace-only reformat (`let cost = …` rejoined to one line). Unrelated to KB-4; ideally reverted to keep the diff clean.
- **URL path segments not encoded** — `project_id` and `doc_id` are interpolated into path segments without percent-encoding. Low risk since these are UUIDs in practice, but a TODO comment would help document the assumption.

PR #1302, opened 2026-06-16
feat(knowledge): KB-5 knowledgebase UI

geoffjay:
Add a **Knowledge** section to the React UI: project picker, virtual document tree, and a CodeMirror 6 markdown editor with debounced autosave and optimistic-concurrency support.

## What's included

**New types** (`ui/src/types/knowledge.ts`):
- `Document`, `DocumentContent`, `TreeNode` (folder/file discriminated union)
- `CreateDocumentRequest`, `UpdateDocumentRequest`, `DocumentListParams`

**KnowledgeClient** (`ui/src/services/knowledge.ts`):
- Extends `ApiClient` following the `MemoryClient` pattern
- 11 typed methods: `getHealth`, `listDocuments`, `createDocument`, `bulkDeleteDocuments`, `getDocument`, `getDocumentContent`, `updateDocument`, `deleteDocument`, `getTree`
- Singleton `knowledgeClient` exported from `services/index.ts`
- `knowledgeServiceUrl` added to `services/config.ts` (port 17011, runtime/env/default resolution)

**Orchestrator additions** (`types/orchestrator.ts`, `services/orchestrator.ts`):
- `Project` type + `ListProjectsParams`
- `listProjects()` method (calls `GET /projects`)

**Pages & components**:
- `KnowledgebasePage.tsx` — thin wrapper (mirrors `MemoriesPage`)
- `pages/knowledge/KnowledgebaseView.tsx` — two-panel layout: sidebar + editor
- `components/knowledge/ProjectPicker.tsx` — dropdown listing all orchestrator projects
- `components/knowledge/DocumentTree.tsx` — expandable folder/file tree with selection state
- `components/knowledge/DocumentEditor.tsx` — CodeMirror 6 editor (one-dark theme, markdown highlighting, debounced autosave at 1 500 ms, optimistic-concurrency `expected_updated_at` token)
- `components/knowledge/DocumentToolbar.tsx` — New/Delete buttons, `CreateDocumentDialog` modal

**Routing & nav** (`App.tsx`, `Sidebar.tsx`):
- Routes: `/knowledge` and `/knowledge/:projectId/*`
- Sidebar nav item: "Knowledge" with `BookOpen` icon

**Tests** (`ui/src/test/services/knowledge.test.ts`):
- 11 vitest tests covering all client methods, query-param forwarding, error handling (404, 409)

**Dependencies installed** (bun):
- `@codemirror/state`, `@codemirror/view`, `@codemirror/commands`
[...]

PR #1303, opened 2026-06-16
feat(knowledge): KB-6 doctor/gc reconciliation, env docs, E2E tests

geoffjay:
Add the `doctor` reconciliation endpoint and CLI subcommand, document the `AGENTD_KNOWLEDGE_PORT` / `AGENTD_KNOWLEDGE_ROOT` environment variables, and add a comprehensive E2E lifecycle test.

## What's included

**`GET /projects/{project_id}/doctor`** - Read-only reconciliation report: lists DB rows with missing disk files and orphaned disk files with no DB row.

**`POST /projects/{project_id}/doctor[?fix=true]`** - Fix mode: automatically deletes stale DB rows and orphaned disk files, returns count of repaired issues.

**`agent knowledge doctor <project_id> [--fix]`** CLI subcommand - Pretty-prints missing/orphaned lists; with `--fix` shows repair count.

**`KnowledgeClient.doctor()` / `.doctor_fix()`** - Typed HTTP client methods for both endpoints.

**`AGENTD_KNOWLEDGE_PORT` / `AGENTD_KNOWLEDGE_ROOT` docs** - Added env var table to `KnowledgeConfig` doc comment in `crates/common/src/config.rs`.

**Tests** (7 new):
- `test_doctor_clean_project_reports_no_divergences`
- `test_doctor_detects_missing_file`
- `test_doctor_detects_orphaned_file`
- `test_doctor_fix_removes_missing_db_row`
- `test_doctor_fix_removes_orphaned_file`
- `test_full_lifecycle_e2e` - create → list → confirm on disk → update → doctor → delete

Closes #1285

### DIFF
--- a/crates/cli/src/commands/knowledge.rs
+++ b/crates/cli/src/commands/knowledge.rs
@@ -14,6 +14,7 @@
 //! - **delete**: Delete a document by ID
 //! - **gc**: Bulk-delete all documents for a project (garbage collect)
 //! - **tree**: Display the virtual folder/file tree for a project
+//! - **doctor**: Reconcile DB rows vs disk files; optionally fix divergences
 //!
 //! # Examples
 //!
@@ -28,6 +29,8 @@
 //! agent knowledge delete <project_id> <doc_id>
 //! agent knowledge gc <project_id>
 //! agent knowledge tree <project_id>
+//! agent knowledge doctor <project_id>
+//! agent knowledge doctor <project_id> --fix
 //! ```
 
 use anyhow::Result;
@@ -193,6 +196,26 @@ pub enum KnowledgeCommand {
         /// Project UUID
         project_id: String,
     },
+
+    /// Reconcile DB rows vs disk files for a project.
+    ///
+    /// Reports missing files (DB rows whose markdown file is absent from disk)
+    /// and orphaned files (disk files with no DB row). Pass `--fix` to
+    /// automatically delete the divergent entries.
+    ///
+    /// # Examples
+    ///
+    /// ```bash
+    /// agent knowledge doctor <project_id>
+    /// agent knowledge doctor <project_id> --fix
+    /// ```
+    Doctor {
+        /// Project UUID
+        project_id: String,
+        /// Automatically fix divergences (delete stale DB rows + orphaned files).
+        #[arg(long)]
+        fix: bool,
+    },
 }
 
 impl KnowledgeCommand {
@@ -251,6 +274,9 @@ impl KnowledgeCommand {
                 cmd_gc(client, project_id, *yes, json).await
             }
             KnowledgeCommand::Tree { project_id } => cmd_tree(client, project_id, json).await,
+            KnowledgeCommand::Doctor { project_id, fix } => {
+                cmd_doctor(client, project_id, *fix, json).await
+            }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


geoffjay (commented):
## Review: KB-6 doctor/gc reconciliation, env docs, E2E tests

**Non-blocking comment** — the KB-6 changes themselves are correct and well-tested. Two suggestions worth addressing before merge, plus a stack note.

---

### Stack note

This PR is at the tip of a 6-PR stack:

> #1296 → #1297 → #1298 → #1299 → **#1300** → **#1302** → #1303

Both #1300 (KB-4, missing status checks + CLI slice panics) and #1302 (KB-5, `BROWSER_SERVICES` gap) currently have `needs-rework`. KB-6 cannot merge until both are resolved. The changes here are independent and reviewable now.

---

### Suggestion 1 — `_total` suffix on a gauge (should fix)

`crates/knowledge/src/storage.rs`, in `doctor()`:

```rust
metrics::gauge!("knowledge_missing_file_total").set(missing_files.len() as f64);
```

The `_total` suffix is reserved by Prometheus convention for **counters** (monotonically increasing values). Using it on a gauge breaks tooling that pattern-matches `_total` metrics for validation, alerting rule generators, and dashboards. A gauge that tracks a current count should use a plain noun or `_count` suffix:

```rust
metrics::gauge!("knowledge_missing_files").set(missing_files.len() as f64);
```

Or if you want to preserve the "how many we've ever seen" intent, consider a counter that increments each time one is detected.

---

### Suggestion 2 — `collect_md_files` follows symlinks (mild safety)

`crates/knowledge/src/storage.rs`, in `collect_md_files_inner`:

```rust
if path.is_dir() {
    collect_md_files_inner(root, &path, out);
}
```

`Path::is_dir()` follows symlinks. A symlinked directory pointing back to an ancestor (circular symlink) will recurse infinitely. In a controlled KB root this is unlikely, but the project root is user-configurable via `AGENTD_KNOWLEDGE_ROOT`. A safe guard:

```rust
if path.is_dir() && !path.is_symlink() {
    collect_md_files_inner(root, &path, out);
}
```

This skips symlinked directories without affecting normal subdirectory traversal.

---

### What looks good

**`DoctorReport` type** is clean and minimal. `Default` is a nice touch (enables `..Default::default()` in tests and avoids boilerplate).

**GET vs POST semantics are clean:**
- `GET /doctor` always runs read-only (`fix = false`, hardcoded).
- `POST /doctor` defaults to `fix = true` — the mutating action lives where it belongs (POST).
- `POST /doctor?fix=false` is technically a no-op POST, but this is harmless and the design doesn't advertise it.

**Client methods** (`doctor` / `doctor_fix`) both check `is_success()` before parsing JSON — consistent with the existing pattern in the file.

**Path separator normalisation** in `collect_md_files_inner`:
```rust
let rel_str = rel.to_string_lossy().replace(std::path::MAIN_SEPARATOR, "/");
```
Correctly handles Windows path separators so `rel_path` values always use `/`, matching the DB storage format.

**Test coverage is thorough:**
- Clean project → no divergences ✓
- Missing file detection and non-fix mode ✓
- Orphan file detection and non-fix mode ✓
- Fix removes stale DB row ✓
- Fix removes orphan disk file ✓
- Full lifecycle E2E: create → list → read → update → doctor → delete ✓

The lifecycle test is particularly good — it validates disk state (`assert!(disk_path.exists())`, `assert_eq!(std::fs::read_to_string(...))`) rather than trusting the DB alone.

**env var table in `KnowledgeConfig` doc comment** is a nice usability improvement.

PR #1304, opened 2026-06-16
feat(knowledge): KB-7 tenant scoping — NULL-row transition policy and gateway docs

geoffjay:
Activates multi-tenancy for the knowledge service now that #1119 has landed. The `organization_id` column and per-query `OptionalTenantId` extraction were already in place from KB-2/KB-3; this issue wires in the NULL-row transition policy and fills the gateway doc gap.

### `crates/knowledge/src/storage.rs`

Added `tenant_read_condition(org)` — a private helper that returns the `Condition` implementing the NULL-row transition policy documented in `agentd-common::tenant`:

```sql
WHERE organization_id = <org> OR organization_id IS NULL
```

The `OR IS NULL` arm ensures legacy documents (created before multi-tenancy) remain readable by every authenticated tenant until an operator runs the backfill command.

Applied to all **read** paths:
- `list_documents` — was `eq(o)`, now uses `tenant_read_condition`
- `find_by_id` (backing `get_document`, `get_document_content`, `update_document`, `delete_document`) — same

**Write paths stay strict:** `bulk_delete_project` with an org scope still uses plain `eq` — an org's GC call should not touch legacy/shared NULL-org documents.

### `crates/core/src/api/gateway.rs`

Added `memory`, `communicate`, and `knowledge` to the route-mapping doc table. The proxy catch-all and `ProxyConfig::from_env()` already included these services (added in #1119); this brings the documentation in line with the implementation.

## Tests (5 new)

| Test | Verifies |
|------|---------|
| `test_null_org_doc_visible_to_any_tenant` | NULL-org doc appears in any tenant's `list` |
| `test_null_org_doc_get_by_any_tenant` | NULL-org doc fetchable by ID from any tenant |
| `test_cross_tenant_isolation` | Org-A docs not visible to org-B and vice-versa |
| `test_tenant_list_mixes_own_and_null_org_docs` | List returns own + NULL docs, excludes other-org docs |
| `test_bulk_delete_org_scoped_leaves_null_org_docs` | GC with org scope does not delete legacy docs |

52 storage tests pass total.

Closes #1286

### DIFF
--- a/crates/core/src/api/gateway.rs
+++ b/crates/core/src/api/gateway.rs
@@ -17,6 +17,9 @@
 //! | `/api/v1/wrap/*`                    | agentd-wrap               |
 //! | `/api/v1/hook/*`                    | agentd-hook               |
 //! | `/api/v1/monitor/*`                 | agentd-monitor            |
+//! | `/api/v1/memory/*`                  | agentd-memory             |
+//! | `/api/v1/communicate/*`             | agentd-communicate        |
+//! | `/api/v1/knowledge/*`               | agentd-knowledge          |
 //!
 //! # WebSocket proxying
 //!
--- a/crates/knowledge/src/storage.rs
+++ b/crates/knowledge/src/storage.rs
@@ -14,8 +14,8 @@
 
 use anyhow::Result;
 use sea_orm::{
-    ActiveModelTrait, ActiveValue::Set, ColumnTrait, DatabaseConnection, EntityTrait, ModelTrait,
-    PaginatorTrait, QueryFilter, QueryOrder, QuerySelect,
+    ActiveModelTrait, ActiveValue::Set, ColumnTrait, Condition, DatabaseConnection, EntityTrait,
+    ModelTrait, PaginatorTrait, QueryFilter, QueryOrder, QuerySelect,
 };
 use std::{
     path::{Path, PathBuf},
@@ -179,8 +179,8 @@ impl KnowledgeStorage {
     ) -> std::result::Result<PaginatedResponse<Document>, KnowledgeError> {
         let mut base_query =
             document::Entity::find().filter(document::Column::ProjectId.eq(project_id));
-        if let Some(o) = org {
-            base_query = base_query.filter(document::Column::OrganizationId.eq(o));
+        if let Some(cond) = tenant_read_condition(org) {
+            base_query = base_query.filter(cond);
         }
         let base_query = base_query.order_by_asc(document::Column::RelPath);
 
@@ -436,8 +436,8 @@ impl KnowledgeStorage {
     ) -> std::result::Result<document::Model, KnowledgeError> {
         let mut query = document::Entity::find_by_id(doc_id.to_string())
             .filter(document::Column::ProjectId.eq(project_id));
-        if let Some(o) = org {
-            query = query.filter(document::Column::OrganizationId.eq(o));
+        if let Some(cond) = tenant_read_condition(org) {
+            query = query.filter(cond);
         }
         query
             .one(&self.db)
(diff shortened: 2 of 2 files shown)


geoffjay (commented):
## Review: KB-7 tenant scoping — NULL-row transition policy and gateway docs

**Non-blocking comment** — the tenant scoping implementation is correct and well-tested; three observations for awareness before this stack merges.

---

### Stack note

This PR sits at the tip of the knowledge epic:

> #1297 → #1300 → #1302 → #1303 → **#1304**

PR #1300 (KB-4) and PR #1302 (KB-5) currently carry `needs-rework`. KB-7 cannot merge until the full stack clears. The KB-7 changes are independently correct and can be reviewed in isolation.

---

### Observation 1 — Write-path behaviour for NULL-org docs is undocumented policy

`tenant_read_condition()` is applied to both `list_documents` and `find_by_id_internal`. Since all mutations (update, delete) go through `find_by_id_internal` to locate the row first, a tenant scoped to `Some("org-a")` can **update and delete** NULL-org legacy documents — not just read them.

The doc comment on `tenant_read_condition` says:

> rows created before multi-tenancy have a `NULL` `organization_id` and must remain **readable** by every authenticated tenant until an operator runs the backfill command.

The implementation, however, also makes them writable. This could be intentional (any tenant can self-service backfill a NULL-org doc into their org), but it also means one tenant could delete a legacy document that every other tenant depends on.

No test currently covers: "what happens when org-a calls `delete_document` on a NULL-org document?"

Suggested fix — either:
- Add a separate `tenant_write_condition` that uses strict equality only (`organization_id = X`, no `OR IS NULL`), so writes are strictly isolated while reads are permissive during the transition.
- Or, document explicitly in `tenant_read_condition`'s doc comment that it intentionally governs both reads and writes during the transition window, and add a test that asserts this behaviour.

---

### Observation 2 — Metric gauge still has `_total` suffix (KB-6 suggestion unaddressed)

The KB-6 review flagged this; it is still present in KB-7:

```rust
// crates/knowledge/src/storage.rs, inside doctor()
metrics::gauge!("knowledge_missing_file_total").set(missing_files.len() as f64);
```

`_total` is a Prometheus-reserved suffix for monotonically increasing counters. A gauge tracking a current snapshot count should use a plain noun or `_count` suffix — e.g. `knowledge_missing_files`. Tooling that validates metric naming (alerting rule generators, Prometheus lint) will reject or misclassify this metric.

---

### Observation 3 — Symlink traversal still unguarded (KB-6 suggestion unaddressed)

Also raised in KB-6 and unchanged:

```rust
if path.is_dir() {
    collect_md_files_inner(root, &path, out);
}
```

`Path::is_dir()` follows symlinks. A circular symlink under a user-configurable `AGENTD_KNOWLEDGE_ROOT` would cause a stack overflow. The guard is a one-liner:

```rust
if path.is_dir() && !path.is_symlink() {
    collect_md_files_inner(root, &path, out);
}
```

---

### What looks good

**`tenant_read_condition()` is clean.** `Condition::any()` builds the SQL `OR` idiomatically, and returning `Option<Condition>` keeps the call site honest — no implicit fallthrough.

**Five tenant isolation tests are thorough:**
- `test_null_org_doc_visible_to_any_tenant` and `test_null_org_doc_get_by_any_tenant` — both list and ID-fetch for the NULL-org transition case
- `test_cross_tenant_isolation` — both directions of cross-org access, covering both list and get-by-ID
- `test_tenant_list_mixes_own_and_null_org_docs` — the mixed scenario with explicit `contains`/`not-contains` path assertions
- `test_bulk_delete_org_scoped_leaves_null_org_docs` — GC respects org boundaries; the unscoped verification (`list_documents(..., None, ...)`) confirms NULL-org doc survival cleanly

**UI fixes address all KB-5 review feedback:**
- `BROWSER_SERVICES` gap fixed — `"knowledge"` added to the constant, the port match arm, and the test assertion in `crates/ui/src/runtime_config.rs` — the runtime URL will now resolve correctly on deployed instances
- Dead `project` state removed from `KnowledgebaseView`
- Unused `splat` destructure removed
- `ProjectPicker` `onError` prop wires fetch errors up to the parent's `setError` rather than silently swallowing them

**Gateway routing table** is now complete for `memory`, `communicate`, and `knowledge`.

---

Observations 2 and 3 were raised in the KB-6 review and remain unaddressed here. If they are not fixed before the stack merges, they should be tracked as follow-up issues.

PR #1319, opened 2026-06-19
feat: add project entity, migration, and storage to core crate

geoffjay:
Adds the canonical `projects` entity to the core crate following the existing `organizations` pattern. Includes SeaORM entity, database migration with unique name index and org_id index, and a full storage layer with CRUD, tenant-scoped listing (with NULL-row legacy compatibility), and pagination.

Closes #1308

### DIFF
--- a/crates/core/src/api.rs
+++ b/crates/core/src/api.rs
@@ -22,6 +22,11 @@
 //! - `GET  /api/v1/organizations/{id}/members`   — list members
 //! - `POST /api/v1/organizations/{id}/members`   — add member (owners only)
 //! - `DELETE /api/v1/organizations/{id}/members/{uid}` — remove member (owners only)
+//! - `POST /api/v1/projects`                     — create project
+//! - `GET  /api/v1/projects`                     — list projects (tenant-scoped via X-Tenant-ID)
+//! - `GET  /api/v1/projects/{id}`                — get project
+//! - `PUT  /api/v1/projects/{id}`                — update project
+//! - `DELETE /api/v1/projects/{id}`              — delete project
 //! - `GET  /api/v1/health`                       — aggregate downstream health check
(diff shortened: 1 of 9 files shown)


geoffjay (commented):
Stack note: This PR is stacked on epic-1306 (PR #1318, currently open). The implementation is ready but must wait for the parent to land first. No restack needed.

## Summary

Clean implementation that faithfully follows the organizations pattern. All 12 tests pass locally. Migration is idempotent, error handling is consistent with the rest of the crate, no unwraps on fallible paths in production code.

## What looks good

- Entity (entity/project.rs): mirrors organization.rs structure; #[sea_orm(unique)] on name is consistent with the migration's unique index.
- Migration (m20260616_000004_create_projects_table.rs): idempotent (if_not_exists() on both table and indexes); down() correctly drops the table (indexes drop automatically). Matches the pattern of other migrations in this crate.
- Storage (project_storage.rs): CRUD methods mirror organization_storage.rs exactly. checked_div for page calculation is the established pattern. list_org NULL-row legacy inclusion is well-documented and the test correctly asserts expected behavior.
- Storage::projects() accessor in storage.rs follows the existing organizations() / memberships() pattern.
- Tests: 12/12 pass; covers create, read-by-id, read-by-name, list, list_org (both branches), paginated list, update, delete, and the unique-name constraint.

## Non-blocking observations

1. list_org(None) returns the full global project set - this is documented, but consider adding a note that callers building HTTP endpoints must not expose this variant without an admin guard.

2. update() cannot reassign organization_id - there is no way via the storage layer to move a legacy NULL-org project into a tenant org. Worth a follow-up issue if the backfill path is not already tracked.

3. Timing sensitivity in test_list_ordered_by_created_at_desc - RFC3339 nanosecond precision makes this reliable in practice, but worth watching if CI clock resolution is coarser.

None of these block merge. LGTM.

PR #1320, opened 2026-06-19
feat: add project CRUD API routes to core crate

geoffjay:
Exposes `/api/v1/projects` as a native core HTTP route with full CRUD. Uses `OptionalTenantId` for org scoping on list and create; handles UNIQUE violations as 409, not-found as 404, and empty-name as 400. Includes 13 handler-level tests covering all happy and error paths.

Closes #1309

### DIFF
--- a/crates/core/src/api.rs
+++ b/crates/core/src/api.rs
@@ -22,6 +22,11 @@
 //! - `GET  /api/v1/organizations/{id}/members`   — list members
 //! - `POST /api/v1/organizations/{id}/members`   — add member (owners only)
 //! - `DELETE /api/v1/organizations/{id}/members/{uid}` — remove member (owners only)
+//! - `POST /api/v1/projects`                     — create project
+//! - `GET  /api/v1/projects`                     — list projects (tenant-scoped via X-Tenant-ID)
+//! - `GET  /api/v1/projects/{id}`                — get project
+//! - `PUT  /api/v1/projects/{id}`                — update project
+//! - `DELETE /api/v1/projects/{id}`              — delete project
 //! - `GET  /api/v1/health`                       — aggregate downstream health check
 //! - `ANY  /api/v1/{service}/*`                  — proxy to downstream service
 
@@ -29,6 +34,7 @@ pub mod admin;
 pub mod auth;
 pub mod gateway;
 pub mod organizations;
+pub mod projects;
 pub mod users;
 
 use axum::{routing::get, Json, Router};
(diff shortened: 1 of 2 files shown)


geoffjay (commented):
## Review: feat: add project CRUD API routes to core crate

**Stack note:** This PR is stacked on issue-1308 (PR #1319, open), which is itself stacked on epic-1306 (PR #1318, open). Both parents need to land first. No restack needed.

---

### Blocking: No authentication on any endpoint

Every handler in this file is missing the `AuthUser` extractor that all organization handlers carry. As written, all five endpoints are accessible without a JWT token.

Compare the organization pattern:
```rust
// organizations.rs - requires auth on every handler
async fn create_org_handler(
    State(state): State<AppState>,
    auth: AuthUser,          // required
    Json(body): Json<CreateOrganizationRequest>,
) -> Result<impl IntoResponse, ApiError>
```

vs. the project handlers:
```rust
// projects.rs - no auth extractor at all
async fn create_project_handler(
    OptionalTenantId(org_id): OptionalTenantId,
    State(state): State<AppState>,
    Json(body): Json<CreateProjectRequest>,
) -> Result<impl IntoResponse, ApiError>
```

There is no router-level auth middleware in `create_router_with_proxy` -- each module is responsible for requiring `AuthUser` on its handlers. Without it:

- Any unauthenticated caller can create, read, update, and delete projects
- `GET /api/v1/projects` with no `X-Tenant-ID` header calls `list_org(None)`, which returns the full global project set -- unauthenticated enumeration of all tenants' data

The tests confirm this: all 13 pass without a single Authorization header, which accurately reflects the (broken) behavior. Those tests should be asserting 401 without auth and 200/201/etc. with auth.

**Fix:** Add `auth: AuthUser` to each handler signature. It can remain unused if per-project authorization rules are deferred to a later PR, but the authentication gate must be there. At minimum, mutating operations (create, update, delete) should enforce it.

---

### Non-blocking observations

**1. Empty name not validated on update** (update_project_handler, line ~195)

`create_project_handler` rejects `name.trim().is_empty()` with a 400. `update_project_handler` has no equivalent check -- `{"name": "   "}` will store a whitespace-only name. Add the same trim-and-check guard before delegating to `storage.update()`.

**2. In-memory pagination in list_projects_handler** (line ~123)

The handler fetches all matching rows via `list_org()` then does `.skip().take()` in memory. `total` is correct (pre-slice count) but loading unbounded rows to page them is a scaling cliff. The storage layer has `list_paginated()` but it lacks tenant-filter support. Worth a follow-up `list_org_paginated(limit, offset, org_id)` method -- flagging so it gets tracked rather than forgotten.

**3. UNIQUE detection via string matching**

`e.to_string().contains("UNIQUE")` is fragile across databases (SQLite vs Postgres error messages differ). This was noted in the storage PR review as well. Low priority while the project is SQLite-only, but worth a typed error check eventually.

PR #1321, opened 2026-06-19
feat: add project CRUD API routes to core crate

geoffjay:
Exposes `/api/v1/projects` as a native core HTTP route with full CRUD. Uses `OptionalTenantId` for org scoping on list and create; handles UNIQUE violations as 409, not-found as 404, and empty-name as 400. All five handlers require `AuthUser` for authentication. Includes 19 tests covering happy paths, 401 unauthenticated cases, and error paths.

Closes #1309

### DIFF
--- a/crates/core/src/api.rs
+++ b/crates/core/src/api.rs
@@ -22,6 +22,11 @@
 //! - `GET  /api/v1/organizations/{id}/members`   — list members
 //! - `POST /api/v1/organizations/{id}/members`   — add member (owners only)
 //! - `DELETE /api/v1/organizations/{id}/members/{uid}` — remove member (owners only)
+//! - `POST /api/v1/projects`                     — create project
+//! - `GET  /api/v1/projects`                     — list projects (tenant-scoped via X-Tenant-ID)
+//! - `GET  /api/v1/projects/{id}`                — get project
+//! - `PUT  /api/v1/projects/{id}`                — update project
+//! - `DELETE /api/v1/projects/{id}`              — delete project
 //! - `GET  /api/v1/health`                       — aggregate downstream health check
 //! - `ANY  /api/v1/{service}/*`                  — proxy to downstream service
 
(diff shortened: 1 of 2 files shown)


geoffjay (commented):
## Review: feat: add project CRUD API routes to core crate (rework of #1320)

**Stack note:** Stacked on issue-1308 (PR #1319, open) → epic-1306 (PR #1318, open). No restack needed. Both parents must land before this merges.

---

### Summary

All blocking issues from the #1320 review are addressed. 19/19 tests pass locally (the pre-existing `test_e2e_register_proxy_and_tenant_isolation` failure requires a live proxy and is unrelated to this PR).

### What changed from #1320

**Blocking issue resolved:** `AuthUser` is now present on all five handlers. Unauthenticated requests correctly return 401 for every endpoint, verified by the five new `*_unauthenticated` tests.

**Non-blocking issues also addressed:**
- `update_project_handler` now validates empty/whitespace-only names with the same `trim().is_empty()` guard as create
- `list_projects_handler` doc comment notes the in-memory pagination limitation and the follow-up path (`list_org_paginated`)
- All 13 former tests are updated to pass auth tokens via a shared `register()` helper, matching the organization test style

### Residual notes (non-blocking, both were flagged as low-priority in #1320)

- **UNIQUE detection via string matching** — `e.to_string().contains("UNIQUE")` remains; still acceptable for SQLite-only, worth a typed error check if/when Postgres support is added
- **In-memory pagination** — documented in the handler; the follow-up `list_org_paginated` storage method should be tracked as its own issue before this service handles large project sets

### One observation for the future

`_auth: AuthUser` enforces authentication but performs no authorization — any authenticated user can modify any project regardless of org membership. That is the right split for this PR (storage layer doesn't have the hooks yet), but it should be tracked as a follow-up once per-project membership or ownership semantics are defined.

LGTM.

PR #1324, opened 2026-06-19
feat: add backfill-projects admin CLI command

geoffjay:
Adds the `agent admin backfill-projects` command to copy project rows from orchestrator's SQLite database into core's SQLite database, preserving UUIDs.

**Implemented:**
- `BackfillProjects` variant added to `AdminCommand` enum with `--dry-run` and optional `--org-id` flags
- Inner `run_backfill_projects` function operates on open DB connections for testability
- Reads all rows from orchestrator's `projects` table (with optional org-id filter including NULL rows)
- Inserts into core's `projects` table preserving original UUIDs; skips existing rows by id (idempotent)
- Human-readable output with colored icons and count summary; `--json` output as array of `{id, name, status, error?}`
- Dry-run mode reports `would_insert` status without writing
- 6 new tests: empty org-id rejection, first-run inserts, idempotency, dry-run no-op, org-id filter with NULL inclusion

Closes #1310

### DIFF
--- a/crates/cli/src/commands/admin.rs
+++ b/crates/cli/src/commands/admin.rs
@@ -8,6 +8,8 @@
 //!
 //! - **backfill-tenant** — Assign an `organization_id` to all existing rows
 //!   that have a NULL value, backfilling legacy unscoped data.
+//! - **backfill-projects** — Copy project rows from orchestrator's database
+//!   into core's database, preserving UUIDs.
 //!
 //! # Usage
 //!
@@ -17,14 +19,22 @@
 //!
 //! # Dry run to preview affected row counts without modifying data
 //! agent admin backfill-tenant --org-id acme-corp --dry-run
+//!
+//! # Copy all orchestrator projects into core
+//! agent admin backfill-projects
+//!
+//! # Dry-run scoped to one organization
+//! agent admin backfill-projects --org-id acme-corp --dry-run
 //! ```
+
+use std::collections::HashSet;
 
 use agentd_common::storage::{create_connection, get_db_path};
 use agentd_install::migrate::DB_SERVICES;
 use anyhow::{Context, Result};
 use clap::Subcommand;
 use colored::*;
-use sea_orm::{ConnectionTrait, Statement, Value};
+use sea_orm::{ConnectionTrait, FromQueryResult, Statement, Value};
 
 /// Admin subcommands.
 #[derive(Debug, Subcommand)]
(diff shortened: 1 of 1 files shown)


crates/cli/src/commands/admin.rs
```diff
@@ -239,6 +281,262 @@ async fn backfill_tenant(org_id: &str, dry_run: bool, json: bool) -> Result<()>
...
+            ProjectBackfillStatus::WouldInsert => {
+                n_would_insert += 1;
+                ("🔍", "would insert")
+            }
+            ProjectBackfillStatus::Error => {
+                n_errors += 1;
+                ("❌", "error")
+            }
+        };
+        println!("  {} {:<40} {}", icon, r.name.bright_black(), verb.green().bold());
```
geoffjay: `verb.green().bold()` is applied to every status including `Error`. This prints the word \"error\" in green — the same colour as a success — which will mislead operators reading migration output. Branch on status so the error arm uses `.red()`:

```rust
let colored_verb = match r.status {
    ProjectBackfillStatus::Error => verb.red().bold(),
    _ => verb.green().bold(),
};
println!(\"  {} {:<40} {}\", icon, r.name.bright_black(), colored_verb);
```

geoffjay (commented):
## Review: feat: add backfill-projects admin CLI command

**Stack position:** PR #1324 is stacked on `epic-1310` (PR #1323, open). Review that parent first if you want the full picture of the migration; this diff is clean in isolation. Note also that the grandparent `feature/projects-to-core` shows `(needs restack)` in git-spice — once that resolves, the whole stack will need an upstack restack.

---

### Blocking — error verb displayed in green (line 475 of admin.rs)

In `output_backfill_projects_results`, a single `.green().bold()` is applied to `verb` for **all** statuses:

```rust
println!("  {} {:<40} {}", icon, r.name.bright_black(), verb.green().bold());
```

When `r.status == ProjectBackfillStatus::Error`, this prints the word "error" in green — the same colour as a successful insert. An operator scanning terminal output after a live migration will see green text everywhere and may not register the failure. The error detail one line below is red, but the per-row status verb should match. Fix by branching on status:

```rust
let colored_verb = match r.status {
    ProjectBackfillStatus::Error => verb.red().bold(),
    _ => verb.green().bold(),
};
println!("  {} {:<40} {}", icon, r.name.bright_black(), colored_verb);
```

---

### Non-blocking suggestions

**Test count discrepancy.** The PR description says "6 new tests" but the diff contains 5 `#[tokio::test]` functions for `backfill_projects`. The natural missing case is an insert that fails due to a name collision (a project with the same `name` but a different `id` already exists in core). The `ProjectBackfillStatus::Error` arm is otherwise untested. Worth adding:

```rust
#[tokio::test]
async fn backfill_projects_name_collision_reports_error() {
    // Seed core with a project using the same name but different UUID already in core,
    // then run backfill. Expect Error status, not a panic.
}
```

**Test helper uses string interpolation for SQL (lines 700–705).** `seed_projects` constructs its INSERT via `format!` with inline values — would break silently on a name containing a single quote, and is inconsistent with the parameterised queries in production paths. Using hardcoded literal values with `execute_unprepared` is fine; avoid the `format!`-plus-variable-data pattern even in test helpers.

---

### What is done well

- Clean separation of I/O (`backfill_projects`) from logic (`run_backfill_projects`) — the inner function is easy to test and the boundary is clearly documented.
- Idempotency via pre-loaded ID set is correct and efficient for migration volumes.
- The org-id filter correctly includes NULL rows (legacy data) — and that behaviour is verified by a dedicated test.
- Error propagation uses `anyhow::Context` throughout; no bare `unwrap` on fallible production paths.
- The empty-`--org-id` guard fires before any database I/O — good fail-fast behaviour.

PR #1329, opened 2026-06-21
feat: repoint MCP tools and UI project calls to core

geoffjay:
Update MCP `run_list_projects` and `run_get_project` tools to target the core service URL (`/api/v1/projects`). Add `core_url` to `AgentdMcpConfig` (env: `AGENTD_CORE_URL`) and `AgentdClient`. Create `ui/src/services/core.ts` with a `CoreClient` and update `ProjectPicker` to use it. Remove `agent_count`/`workflow_count` from `ProjectDetail` since core does not compute them.

Closes #1312

### DIFF
--- a/crates/common/src/config.rs
+++ b/crates/common/src/config.rs
@@ -455,6 +455,8 @@ impl Default for CorePamConfig {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(default)]
 pub struct McpConfig {
+    /// Core service URL (API gateway). Defaults to `"http://localhost:17000"`.
+    pub core_url: String,
     /// Orchestrator service URL. Defaults to `"http://localhost:17006"`.
     pub orchestrator_url: String,
     /// Notify service URL. Defaults to `"http://localhost:17004"`.
@@ -478,6 +480,7 @@ pub struct McpConfig {
 impl Default for McpConfig {
     fn default() -> Self {
         Self {
+            core_url: "http://localhost:17000".to_string(),
             orchestrator_url: "http://localhost:17006".to_string(),
             notify_url: "http://localhost:17004".to_string(),
             ask_url: "http://localhost:17001".to_string(),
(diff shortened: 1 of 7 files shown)


geoffjay (commented):
## Review: feat: repoint MCP tools and UI project calls to core

**Stack position:** This PR (`issue-1312`) is stacked on `issue-1311` (PR #1328). PR #1328 was just given a `needs-rework` label in its own review. This PR's code is clean and correct, but it cannot merge until #1328 is addressed and its fixes propagate up the stack.

---

### Non-blocking: stale JSDoc comment in `ProjectPicker.tsx`

The file-level comment at lines 1–4 still describes the component as fetching from the orchestrator:

```
/**
 * ProjectPicker — dropdown that lists all projects from the orchestrator and
 * lets the user select one as the active knowledge context.
 */
```

Should be updated to reference core now that the import switched from `orchestratorClient` to `coreClient`. Easy one-liner fix.

---

### Non-blocking: no integration test coverage for the new core routing

`tests/common/mod.rs` correctly adds `core_url: "http://127.0.0.1:1".to_string(), // unused` to the test client so it compiles, but `run_list_projects` and `run_get_project` have no integration tests against a mock core server. This gap predates this PR — the functions weren't tested before either — so it's not a regression, but a follow-up test using a `mock_core_server()` modelled on the existing `mock_orch_server()` would be worthwhile.

---

### Looks good

- **ENV var naming is consistent.** `AGENTD_CORE_URL` in the MCP crate and `AGENTD_MCP_CORE_URL` in common config follow the exact same pattern used for all other services (`AGENTD_ORCHESTRATOR_URL` / `AGENTD_MCP_ORCHESTRATOR_URL`, etc.). No inconsistency.
- **URL construction is correct.** `CoreClient` sets `baseUrl = serviceConfig.coreServiceUrl` (default `http://localhost:17000`) and each method appends `/api/v1/projects`. `ApiClient` constructs the final URL as `${this.baseUrl}${path}` → `http://localhost:17000/api/v1/projects`. ✓
- **`withAuth` pattern followed correctly.** `CoreClient` matches the established pattern used by every other service client (`ask`, `communicate`, `knowledge`, `monitor`, `memory`). ✓
- **`ProjectDetail` cleanup is correct.** After removing `agent_count`/`workflow_count`, the struct has `id`, `name`, `description`, `created_at`, `updated_at` — which exactly matches the core `ProjectResponse` fields. Serde deserialization will work without any extra attributes needed. ✓
- **MCP config wiring is complete.** `AgentdMcpConfig::from_env()`, `ValidateConfig`, `Default`, the env-clear lists in all three tests, and `test_validate_bad_url_fails` are all updated consistently. ✓
- **`orchestratorClient.listProjects` is now dead code in the UI** (line 324 of `orchestrator.ts` has no remaining callers after this PR). Not a problem for this PR but worth a follow-up cleanup.

PR #1330, opened 2026-06-22
feat: add project migration tests and update documentation

geoffjay:
Implements the Phase 5 cleanup for the project entity migration (#1305) — integration tests, association endpoint tests, CLI routing tests, and documentation updates.

**Core handler tests** (`crates/core/src/api/projects.rs`):
- Added `test_list_projects_with_tenant_header_scopes_results` — verifies `X-Tenant-ID` header scoping on `GET /api/v1/projects` (org-filtered results include matching + NULL-org rows; no header returns all)

**Orchestrator association tests** (new `crates/orchestrator/tests/project_associations_http.rs`):
- 9 tests covering `POST`/`DELETE` agent and workflow associations, 404 on missing agent/workflow, and documented behavior that the orchestrator does not validate project IDs against core

**CLI integration tests** (`crates/cli/tests/integration_test.rs`):
- New `project_tests` module with 6 mockito tests verifying: CRUD methods route to core URL, association methods route to orchestrator URL, `delete_project` issues association pre-checks before deleting from core, and `delete_project` returns an error when agents are still associated

**Documentation**:
- `README.md`: add `agentd-core` to port table (17000/7000), update core description to mention project management, add project management CLI entry, add v0.15.0 project entity migration status section with operator upgrade note
- `docs/storage.md`: add service-to-table ownership table noting `projects` moved from orchestrator to core in v0.15.0

Closes #1314

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ A modular daemon system for managing AI agents, notifications, interactive quest
 **agentd** is a suite of services and tools designed to orchestrate AI agents and provide intelligent, context-aware notifications and interactions. It consists of:
 
 - **agent** - Command-line interface for interacting with all services
-- **agentd-core** - Core service providing user and organization management
+- **agentd-core** - Core service providing user, organization, and project management
 - **agentd-orchestrator** - Agent lifecycle management, WebSocket SDK server, workflow scheduler, and tool policy enforcement
 - **agentd-notify** - Notification service with REST API and SQLite storage
 - **agentd-ask** - Interactive question service with tmux integration
@@ -97,6 +97,7 @@ agent teardown .agentd/               # delete in reverse order
 - **Declarative templates** - `agent apply` / `agent teardown` for YAML-based agent and workflow management
 - **Agent management** - create, list, get, delete, attach, send-message, stream
 - **Workflow management** - create, list, get, update, delete, history, validate-template
+- **Project management** - create, list, show, update, delete; add/remove agent and workflow associations
 - **Tool policies** - get-policy, set-policy, `--tool-policy` flag on create-agent
 - **Approval management** - list-approvals, approve, deny (for RequireApproval policy)
 - **Health monitoring** - `agent status` checks all services concurrently; per-service `health` commands
@@ -362,12 +363,13 @@ For the complete configuration reference including all environment variables, da
 
 | Service | Dev Port | Prod Port | Description |
 |---------|----------|-----------|-------------|
+| agentd-core | 17000 | 7000 | Core API (users, organizations, projects) |
 | agentd-ask | 17001 | 7001 | Interactive question service |
 | agentd-hook | 17002 | 7002 | Shell hook integration |
 | agentd-monitor | 17003 | 7003 | System monitoring |
 | agentd-notify | 17004 | 7004 | Notification service |
 | agentd-wrap | 17005 | 7005 | Tmux session management |
-| agentd-orchestrator | 17006 | 7006 | Agent orchestration |
+| agentd-orchestrator | 17006 | 7006 | Agent orchestration and project associations |
 | agentd-index | 17012 | 17012 | Semantic code search and indexing |
 | agentd-mcp | - | - | MCP server (stdio transport, no HTTP port) |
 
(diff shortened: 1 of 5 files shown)


geoffjay (commented):
## Review: feat: add project migration tests and update documentation

**Stack position:** `epic-1314` is stacked directly on `feature/projects-to-core` (#1325). Two commits sit above the base in this branch — both scoped to tests and docs. The implementation (including `with_core_url`, the `delete_project` association pre-check, and the core-routing helpers) is already present in the base, so the tests compile and exercise live code. This is the correct final PR to merge in the series.

**Note on PR #1328:** The `delete_project` association pre-check IS implemented correctly here (`list_project_agents` then `list_project_workflows`, bail early if either is non-zero, then `delete_core`). The `needs-rework` on #1328 can be narrowed to just fixing the now-stale docstring — the safety concern is resolved.

---

### Approved — non-blocking suggestions below

---

### Non-blocking: no test for `delete_project` blocked by *workflows* (only agents case is covered)

`test_delete_project_blocked_when_agents_associated` exercises the agents guard. There is no symmetric test for when ONLY workflows are still associated. The implementation calls `list_project_agents` first and short-circuits on agents > 0, so the workflow guard is never reached in that test — but the workflow guard itself is untested. A minimal complement:

```rust
#[tokio::test]
async fn test_delete_project_blocked_when_workflows_associated() {
    let core_server = Server::new_async().await;
    let mut orch_server = Server::new_async().await;
    let project_id = Uuid::new_v4();
    let wf_id = Uuid::new_v4();

    // Agents: none
    orch_server
        .mock("GET", format!("/projects/{project_id}/agents").as_str())
        .with_status(200)
        .with_body(json!({ "items": [], "total": 0, "limit": 50, "offset": 0 }).to_string())
        .with_header("content-type", "application/json")
        .create_async().await;

    // Workflows: one remaining
    orch_server
        .mock("GET", format!("/projects/{project_id}/workflows").as_str())
        .with_status(200)
        .with_body(json!({
            "items": [{ "id": wf_id, "name": "busy-wf" }],
            "total": 1, "limit": 50, "offset": 0
        }).to_string())
        .with_header("content-type", "application/json")
        .create_async().await;

    let client = OrchestratorClient::new(orch_server.url()).with_core_url(core_server.url());
    let result = client.delete_project(&project_id).await;
    assert!(result.is_err());
    let msg = result.unwrap_err().to_string();
    assert!(msg.contains("workflow") || msg.contains("associated"), "{msg}");
}
```

---

### Non-blocking: `project_id FK` language in `docs/storage.md` is technically inaccurate

The table says:

```
| `agents` | agentd-orchestrator | `agentd/agent.db` | `project_id` FK points to core `projects.id` |
```

These are separate SQLite databases on separate services — there is no database-enforced foreign key constraint. "FK" implies referential integrity that does not exist here. Consider phrasing it as a logical reference instead:

> `project_id` references core `projects.id` (logical ref — no DB-enforced constraint)

---

### Non-blocking: `agentd-index` is still in the README (pre-existing, out of scope)

`CLAUDE.md` documents that the index crate was removed as an experiment. Lines 29 and 373 in `README.md` still list `agentd-index`. This PR is already editing the port table and service list — a good opportunity to drop those stale rows — but it is not required for approval.

---

### Everything else looks good

- **Core handler test** (`test_list_projects_with_tenant_header_scopes_results`) — correctly seeds 3 projects with different org scoping and asserts that `X-Tenant-ID: org-x` returns 2 (matching + NULL-org rows), while a headerless request returns 3. Matches the `list_org()` storage contract precisely.
- **Orchestrator association tests** (9 tests) — `NullBackend` is comprehensive, `build_app()` is clean and reusable, agent names are unique per test to prevent parallel-test flakiness, and `test_associate_agent_nonexistent_project_still_succeeds` explicitly documents the deliberate lack of cross-service project validation with a clear comment.
- **CLI routing tests** — two-server mockito setup cleanly separates core vs. orchestrator traffic. `test_delete_project_checks_associations_then_deletes_from_core` asserts all three mocks (agents list, workflows list, core DELETE) via `assert_async()`, which is exactly right.
- **Documentation** — port table addition, orchestrator description update, v0.15.0 migration status block, and the `backfill-projects` operator note are all accurate and match the implementation.

